Issue #219, opened 2014-03-07
converter for datetime objects doesn't handle timezones


exarkun:
pymysql.converters.escape_datetime ignores the timezone of the datetime object passed to it.

The result is that two datetime instances that represent the same time but have different timezones end up looking like two different times instead.  For example:

```
>>> pymysql.converters.escape_datetime(datetime.datetime.now(pytz.timezone("US/Eastern")))
"'2014-03-07 11:16:07'"
>>> pymysql.converters.escape_datetime(datetime.datetime.now(pytz.timezone("UTC")))
"'2014-03-07 16:16:09'"
```

Though I only evaluated these two expressions two seconds apart the resulting values appear to be four hours and two seconds apart.

methane:
Application should convert application's datetime with tz to database's (naive) datetime.
If you don't want to do it manually, you should use some wrapper.  Don't use this library directly.

PR #253, opened 2014-07-16
Removing timezone to enable escaping

BradRuderman:
Hi All-
There is an issue here related to #219 where PyMySQL should strip the timezone from the datetime before inserting. Right now it is failing on the insert, since mysql doesn't take timezone info.

Thanks!

### DIFF
--- a/pymysql/converters.py
+++ b/pymysql/converters.py
@@ -89,9 +89,11 @@ def escape_time(obj):
     return escape_str(s)
 
 def escape_datetime(obj):
+    obj = obj.replace(tzinfo=None)
     return escape_str(obj.isoformat(' '))
 
 def escape_date(obj):
+    obj = obj.replace(tzinfo=None)
     return escape_str(obj.isoformat())
 
 def escape_struct_time(obj):


pymysql/converters.py
```diff
@@ -89,9 +89,11 @@ def escape_time(obj):
     return escape_str(s)
 
 def escape_datetime(obj):
+    obj = obj.replace(tzinfo=None)
```
marc-mabe: For MySQL, it's true that there is no timezone support in datetime formats but it handles the timezone as follows:
- `TIMESTAMP` gets stored as UTC and will be converted from/to current `session.time_zone` on INSERT/UPDATE/SELECT
- `DATETIME` doesn't handle time zones at all

Normally you setup the timezone in MySQL on the server once and never change that (hopefully UTC) but you can change the time zone on runtime by `SET @@session.time_zone = 'Europe/Berlin'`.

I think to make it right you should convert the `datetime` object to the time zone provided by `SELECT IF(@@session.time_zone="SYSTEM",@@global.time_zone,@@session.time_zone)` as long as the object has a time zone.

I know this would be slow on doing it each time escape gets called but it could be cached in the connection.

Thoughts ?

methane: I think engineer should know and care about it.
I don't want to do anything magical.

PR #279, opened 2015-01-07
Implement load data local infile command. Resolves #62

### LINKED ISSUE #62: pymysql does not support  "LOAD DATA LOCAL" ?
Originally filed by fla.sam on 2011-04-29T08:58:02

What steps will reproduce the problem?
1. I try to load a local data from a csv format file into remote mysql-server"
LOAD DATA LOCAL INFILE 'a_csv_file' 
            INTO TABLE some_table_name
            FIELDS 
                TERMINATED BY ','
                ENCLOSED BY '"'
            LINES 
                TERMINATED BY '\n'
            IGNORE 1 LINES
            (field1, field2)
" 

What is the expected output? What do you see instead?
  File "/home/sam/workspace/lqjh2/lqjh2_admin_srv/pre_data/load_pre_data.py", line 62, in main
    cursor.execute(sql)
  File "/home/sam/PY_ENV/lq2_ENV/lib/python2.7/site-packages/django/db/backends/util.py", line 15, in execute
    return self.cursor.execute(sql, params)
  File "/home/sam/PY_ENV/lq2_ENV/lib/python2.7/site-packages/django/db/backends/mysql/base.py", line 86, in execute
    return self.cursor.execute(query, args)
  File "/home/sam/PY_ENV/lq2_ENV/lib/python2.7/site-packages/pymysql/cursors.py", line 108, in execute
    self.errorhandler(self, exc, value)
  File "/home/sam/PY_ENV/lq2_ENV/lib/python2.7/site-packages/pymysql/connections.py", line 184, in defaulterrorhandler
    raise errorclass, errorvalue
django.db.utils.DatabaseError: (1148, u'The used command is not allowed with this MySQL version')

but when i switch to "MySQL-python" it's everything fine.

What version of the product are you using? On what operating system?
0.4

wraziens:


pymysql/connections.py
```diff
@@ -15,6 +14,8 @@
 import socket
 import struct
 import sys
+import re
+import traceback
```
methane: Could you sort imports alphabetical order?

wraziens: The imports should now be in alphabetical order.

pymysql/connections.py
```diff
@@ -472,7 +497,7 @@ def __init__(self, host="localhost", user=None, password="",
                  client_flag=0, cursorclass=Cursor, init_command=None,
                  connect_timeout=None, ssl=None, read_default_group=None,
                  compress=None, named_pipe=None, no_delay=False,
-                 autocommit=False, db=None, passwd=None):
+                 autocommit=False, db=None, passwd=None, load_local=False):
```
methane: Could you change option name to `local_infile`?
Options other than DB-API 2.0, I like consistency with MySQL options.
http://dev.mysql.com/doc/refman/5.1/en/server-system-variables.html#sysvar_local_infile

methane: And add it to docstring.

pymysql/connections.py
```diff
@@ -890,6 +919,12 @@ def _execute_command(self, command, sql):
         if isinstance(sql, text_type):
             sql = sql.encode(self.encoding)
 
+        local_file_name = re.search(b'load data local infile \'(?P<file_name>[^\']+)\'', sql, flags=re.IGNORECASE)
+        if local_file_name:
```
methane: I feel this check is too much.
This expression doesn't match when there are TABs or LFs.
Since users should enable `local_infile` option only when they believe server, I'd like to completely remove this check.

pymysql/connections.py
wraziens: Done.

pymysql/connections.py
wraziens: Removed.

pymysql/connections.py
```diff
@@ -1191,4 +1227,70 @@ def _get_descriptions(self):
...
+            with open(self.filename, 'r') as open_file:
+                chunk_size = MAX_PACKET_LEN
+                prelude = b""
+                packet = b""
+                packet_size = 0
+
+                for line in open_file:
+                    line_length = len(line)
+                    format_str = '!{0}s'.format(line_length)
+                    format_str = format_str.encode(self.connection.encoding)
```
methane: Ref http://dev.mysql.com/doc/refman/5.1/en/load-data.html
MySQL uses character_set_database variable or "CHARACTER SET" clause.  Connection encoding should not affect file content.

Open file with 'rb' mode and use loop like:
```python
while True:
    chunk = open_file.read(chunk_size)
    if not chunk:
        break
    packet = struct.pack('<i', len(chunk))[:3] + int2byte(seq_id)
    self.connection._write_bytes(packet)
    seq_id += 1
```

pymysql/connections.py
```diff
@@ -1191,4 +1227,70 @@ def _get_descriptions(self):
...
+
+class LoadLocalFile(object):
+    def __init__(self, filename, connection):
+        self.filename = filename
+        self.connection = connection
+
+    def _print_warnings(self):
+        self.connection._execute_command(COMMAND.COM_QUERY, 'SHOW WARNINGS')
+        warnings = MySQLResult(self.connection)
+        warnings.read()
```
methane: Remove this method.  Utility method like this should implemented by user or separated function.
Since `print()` may fails, protocol implementation should not print anything.

pymysql/connections.py
```diff
@@ -1191,4 +1227,70 @@ def _get_descriptions(self):
...
+
+        except IOError:
+            raise OperationalError(1017, "Can't find file '{0}'".format(self.filename))
+        finally:
+            # send the empty packet to signify we are done sending data
+            packet = struct.pack('<i', 0)[:3] + int2byte(seq_id)
+            self.connection._write_bytes(packet)
+
+            result = MySQLResult(self.connection)
+            result.read()
```
methane: Move this to caller.  Two result object for one query may cause bug.

pymysql/connections.py
```diff
@@ -1100,6 +1131,11 @@ def _read_ok_packet(self, first_packet):
         self.message = ok_packet.message
         self.has_next = ok_packet.has_next
 
+    def _read_load_local_packet(self, first_packet):
+        load_packet = LoadLocalPacketWrapper(first_packet)
+        local_packet = LoadLocalFile(load_packet.filename, self.connection)
+        local_packet.send_data()
+
```
methane: ok_packet = self.connection._read_packet()
if not ok_packet.is_ok_packet():
    raise OperationalError(2014, "Commands Out of Sync")
self._read_ok_packet(ok_packet)

pymysql/connections.py
wraziens: If we remove this then we will not have the same functionality as mysqldb, which I thought was a goal of pymysql, no? The user will have no idea if there were any warnings returned after their file was loaded, which seems a bit dangerous to me. 

Also, out of curiosity, how do you expect print() to fail?

pymysql/connections.py
wraziens: I'm not sure what you are getting at here.. This is not an ok_packet, it has a different format than an ok packet and needs to be handled differently...

methane: I said about adding these code after `local_packet.send()`, as described below.

pymysql/connections.py
methane: stdout may not valid file, or closed.
For example, print() may raise exception on mod_wsgi's default setting. (ref. https://code.google.com/p/modwsgi/wiki/ConfigurationDirectives#WSGIRestrictStdout)

I check how MySQLdb treat warnings.  It check MySQL warnings on cursor, and use `warnings.warn()` to show it.
https://github.com/farcepest/MySQLdb1/blob/d34fac681487541e4be07e6978e0db233faf8252/MySQLdb/cursors.py#L107-L120

pymysql/connections.py
wraziens: Oh, ok. That makes more sense, so instead of getting the response packet in the finally clause of send_data() we will handle it here.

pymysql/connections.py
wraziens: I changed the print statements to warnings.warn.

methane:
Thanks for great job.

John-Nagle:
Great! I'm upgrading from MySQLdb, and need that. 
When will this version be pushed to PyPi?  PyPi is still at 0.6.3, and this is in 0.6.4.

chriscannon:
I also think 0.6.4 should be pushed to PyPi.

John-Nagle:
Er, yes. I didn't find out until the end of a long batch run that LOAD DATA INFILE didn't work.

PR #280, opened 2015-01-11
Added commit method

719Ben:
Without the commit method nothing is saved to the database. Need to add the commit() method in order to be a better example.

### DIFF
--- a/example.py
+++ b/example.py
@@ -17,5 +17,6 @@ print()
 for row in cur:
    print(row)
 
+cur.commit()
 cur.close()
 conn.close()
--- a/example.py
+++ b/example.py
@@ -18,6 +18,6 @@ print()
 for row in cur:
    print(row)
 
-cur.commit()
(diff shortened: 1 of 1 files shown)


719Ben:
While it may not be needed for a simple SELECT statement, there should be an example for code that need the commit() method.

methane:
Is that example runs correctly?
Connection object has `.commit()` method.  But cursor may not have it.

719Ben:
You are correct, it should `.commit()` on the connection object. I have edited the code reflect that.

example.py
```diff
@@ -17,5 +18,6 @@
 for row in cur:
    print(row)
 
-cur.close()
+conn.commit()
+curr.close()
```
methane: s/curr/cur/

methane:
Inserting user in example is bad idea.
We should have more friendly docs and examples.

PR #285, opened 2015-01-27
Repair test_datatypes to support MySQL 5.6 datetime rounding behavior

zzzeek:
per http://dev.mysql.com/doc/refman/5.6/en/fractional-seconds.html, we can observe that sending in a datetime with microseconds > 500000 results in rounding behavior of one second forward, causing this test to fail 50% of the time on such a server, as it is sending in datetime.now() without any truncation of microseconds.  The patch adjusts the test to check for MySQL server version >= 5.6.4 and microseconds > 500000, rounding the value up by one second if so, so that the test can pass in all cases.

### DIFF
--- a/pymysql/tests/base.py
+++ b/pymysql/tests/base.py
@@ -1,6 +1,7 @@
 import os
 import json
 import pymysql
+import re
 try:
     import unittest2 as unittest
 except ImportError:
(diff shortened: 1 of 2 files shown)


methane:
I'd like to remove `datetime.datetime.now()` and use constant value for test.
It may makes the code simple.

zzzeek:
sure thing, though you might want to then add a test to ensure that MySQL's expected rounding behavior on 5.6.4 is preserved when one does in fact pass a datetime with a microseconds > 500000.

zzzeek:
this is dependent on the fact that you're sending `obj.isoformat(' ')` along; if that ever changed to not include microseconds then that behavior would break.  So it should have a test.

methane:
I think we should test PyMySQL encode and decode datetime correctly.
But I think we don't have to test MySQL behavior.
Avoiding such behavior makes running tests against MariaDB or other DBs easy.

Issue #287, opened 2015-01-30
KeyboardInterrupt exception forces finishing of query


mathieulongtin:
When you hit ctrl-C in the middle of a query, the cursor's `__del__` method calls cursor.close(), which tries to finish the current query. So if you hit ctrl-C in the middle of a long query, it tries to grab all the results from that query rather than killing it.

mathieulongtin:
I'm not sure how to handle that. I think maybe removing the __del__ that are all over the place. If someone drops an unfinished cursor, they shouldn't have any guarantee that the connection will keep working. The only thing is the connection should probably send a COM_QUIT before closing the socket just to be polite.

zzzeek:
it's very standard and largely expected that when a Python object is GC'ed, it leaves the program in a stable state and that remaining, non-GC'ed objects continue to work correctly (all DBAPIs do this)   If the connection is still present, it definitely should still be in a working state.   If your program is trapping ctrl-C and is allowing the connection to remain usable, it doesn't have much option here.   OTOH, if your program is exiting out, the connection itself will be closed and there wont be anything left to read by the cursor (I suppose if that's not happening, PyMySQL could replace the use of __del__ with weakref callbacks, which are much more controllable).

Maybe you need to catch KeyboardInterrupt and call connection.close(), if you're wanting the program to catch ctrl-C and continue executing without any delay.

mathieulongtin:
That's the problem, I do catch the exception, but my catcher never gets executed, it's stuck in the close method of the cursor. 

—
Mathieu Longtin
514-803-8977

zzzeek:
and its the `__del__()`?  like, if you refer to the cursor in the KeyboardInterrupt, the reference stays, `__del__()` doesn't get called, and it completes quickly?  this doesn't really sound like a possible thing.  Have any test code illustrating this?

mathieulongtin:
Ok, let me qualify that. This is only an issue with `SSCursor`. Default `Cursor` behaves as expected.

I created some sort of test bed for the various combinations here: https://gist.github.com/mathieulongtin/7ad1227285ebaa35a70e. Just change the SQL to something slow.

If I run the code with just --stream, and I hit ctrl-C after a few seconds, I get this:

```
$ python ./pymysql_ctrlc_testbed.py --stream
Got 1081 rows Caught: KeyboardInterrupt()
Did exception get caught?
Got 1081 rows
[...]

PR #292, opened 2015-02-01
Remove __del__ from Cursor; warn on unconsumed SSCursor results

zzzeek:
This PR adds tests to confirm that things are cleaned up if a Cursor or SSCursor is discarded in the middle of results.   The cleanup routine of MySQLResult handles this already.   A warning is added to Connection if an unbuffered result was left open, and a subsequent execute call is made.

With this patch, the issue of slow program termination reported in #287 is resolved.  

What I am not explicitly testing, but I *think* is OK, is if the statement had multiple result sets.  It looks like the way MySQLResult._finish_unbuffered_query works should handle this as well, but let me know if that's not the case.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,10 @@ Changes
 0.6.4 -Support "LOAD LOCAL INFILE".  Thanks @wraziens
       -Show MySQL warnings after execute query.
       -Fix MySQLError may be wrapped with OperationalError while connectiong. (#274)
+      -SSCursor no longer attempts to expire un-collected rows within __del__,
+       delaying termination of an interrupted program; cleanup of uncollected
+       rows is left to the Connection on next execute, which emits a
+       warning at that time. (#287)
 
 0.6.3 -Fixed multiple result sets with SSCursor.
       -Fixed connection timeout.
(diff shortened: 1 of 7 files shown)


pymysql/cursors.py
```diff
@@ -396,6 +390,7 @@ def fetchall(self):
         Fetch all, as per MySQLdb. Pretty useless for large queries, as
         it is buffered. See fetchall_unbuffered(), if you want an unbuffered
         generator version of this method.
+
         """
```
methane: FYI, empty line at end of docstring recommendation was removed.
https://hg.python.org/peps/rev/a7b2b2d5b1d4

CHANGELOG
```diff
@@ -4,6 +4,10 @@ Changes
 0.6.4 -Support "LOAD LOCAL INFILE".  Thanks @wraziens
       -Show MySQL warnings after execute query.
       -Fix MySQLError may be wrapped with OperationalError while connectiong. (#274)
+      -SSCursor no longer attempts to expire un-collected rows within __del__,
+      delaying termination of an interrupted program; cleanup of uncollected
+      rows is left to the Connection on next execute, which emits a
+      warning at that time. (#287)
```
methane: Add more one space to indent, like followings.

CHANGELOG
```diff
@@ -47,7 +51,7 @@ Changes
       -Removed DeprecationWarnings
       -Ran against the MySQLdb unit tests to check for bugs
       -Added support for client_flag, charset, sql_mode, read_default_file,
-       use_unicode, cursorclass, init_command, and connect_timeout. 
+       use_unicode, cursorclass, init_command, and connect_timeout.
```
zzzeek: this is the only non-related syntax change that it's a real PITA for me to avoid as my editor strips off trailing whitespace.

PR #299, opened 2015-02-02
Protocol parsing optimizations

YorikSar:
I've based this on benchmark from Mike Bayer's mini-benchmark from his [evaluation of PyMySQL] (https://wiki.openstack.org/wiki/PyMySQL_evaluation). It showed that the most time-consuming parts of protocol parsing code are:
* converting rows from protocol representation to Python objects;
* lots of `read()` calls;
* huge amount of method calls overall.
This branch summarizes fixes that optimize all of these aspects. Generally speedup is gained by:
* moving work that can be done before row parsing loop to the top of it;
* avoiding copying buffers around (almost all but most general read() calls are eliminated, bytes object concatenation is avoided);
* reducing amount of calls to `struct.unpack` methods my squashing consecutive calls into one (and replacing them with `struct.unpack_from` to avoid extra data copying).

On my machine this branch showed around 25% speed gain in Mike's benchmark.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -291,23 +291,54 @@ class MysqlPacket(object):
         """
         return self._data[position:(position+length)]
 
+    if PY2:
+        def read_uint8(self):
+            result = ord(self._data[self._position])
+            self._position += 1
+            return result
+    else:
+        def read_uint8(self):
+            result = self._data[self._position]
+            self._position += 1
+            return result
+
+    def read_uint16(self):
+        result = struct.unpack_from('<H', self._data, self._position)[0]
+        self._position += 2
+        return result
+
+    def read_uint24(self):
+        low, high = struct.unpack_from('<HB', self._data, self._position)
+        self._position += 3
+        return low + high << 16
+
+    def read_uint32(self):
+        result = struct.unpack_from('<I', self._data, self._position)[0]
+        self._position += 4
+        return result
+
+    def read_uint64(self):
+        result = struct.unpack_from('<Q', self._data, self._position)[0]
+        self._position += 8
+        return result
+
     def read_length_encoded_integer(self):
         """Read a 'Length Coded Binary' number from the data buffer.
 
         Length coded numbers can be anywhere from 1 to 9 bytes depending
         on the value of the first byte.
         """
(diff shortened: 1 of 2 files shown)


zzzeek:
good effort, lots of test failures so far.   Did you try getting all tests to pass locally?

YorikSar:
@zzzeek, yeah, forgot to add last piece to last commit. Should be fixed now.

pymysql/connections.py
```diff
@@ -1228,11 +1241,31 @@ def _read_row_from_packet(self, packet):
...
             self.fields.append(field)
             description.append(field.description())
+            field_type = field.type_code
+            if use_unicode:
+                if field_type in TEXT_TYPES:
+                    charset = charset_by_id(field.charsetnr)
+                    if charset.is_binary:
+                        encoding = None
+                    else:
+                        # TEXTs with charset=binary means BINARY types.
```
methane: This comment line should be after `if charset.is_binary:`.

methane:
Wow, great job!  Thank you.

methane:
UNSIGNED_CHAR_LENGTH, UNSIGNED_SHORT_LENGTH, UNSIGNED_INT24_LENGTH and UNSIGNED_INT64_LENGTH are not used anymore.

Although it doesn't have `_` prefix, I don't think they are public API.  Could you remove them?

methane:
Also unpack_uint16, unpack_int24, unpack_int32 and unpack_int64

YorikSar:
I have a question. I used `ascii` encoding to decode non-binary non-text data before passing it to converter. Is it OK or should I use `latin1` or `DEFAULT_CHARSET`?

methane:
I think yes.
All connection encoding should be ascii compatible (e.g. utf-8 or koi8r), and all types other than text and binary (e.g. DATETIME or INTEGER)should be represented in ascii.

methane:
LGTM.
@zzzeek Please merge if you are OK too.

zzzeek:
it looks good, I'm running the entire SQLAlchemy test suite against it, there are four failures but I get the same ones against master as well, so they were not introduced here :).  

thanks for the quick turnaround!

PR #300, opened 2015-02-02
fix issue 288: executemany now works with "insert ... on duplicate update" 

mathieulongtin:
Instead of falling back to executing each statement separately when "on duplicate update" is there, it builds a batch call.

### DIFF
--- a/pymysql/cursors.py
+++ b/pymysql/cursors.py
@@ -147,7 +147,7 @@ class Cursor(object):
         if m:
             q_prefix = m.group(1)
             q_values = m.group(2).rstrip()
-            q_postfix = m.group(3) or ''
+            q_postfix = bytearray(m.group(3) or '')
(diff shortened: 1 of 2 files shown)


pymysql/cursors.py
```diff
@@ -145,17 +145,18 @@ def executemany(self, query, args):
 
         m = RE_INSERT_VALUES.match(query)
         if m:
-            q_values = m.group(1).rstrip()
+            q_prefix = m.group(1)
+            q_values = m.group(2).rstrip()
+            q_postfix = bytearray(m.group(3) or '')
```
methane: prefix is encoded in `_do_execute_many()`. postfix should be encoded in same way.

mathieulongtin: Thanks for the tip. Just done that and it passes python 3.x tests.

PR #305, opened 2015-02-05
Support custom mapping for connection's escape method

boglin:
MySQLdb provides a second argument to connection's escape method where an 'encoders' dict can be supplied.

For compatibility with existing code could this be added to PyMySQL?

From MySQLdb, _mysql.c:
```
static char _mysql_escape__doc__[] =
"escape(obj, dict) -- escape any special characters in object obj\n\
using mapping dict to provide quoting functions for each type.\n\
Returns a SQL literal string.";
```

Proposed test case:

```
def test_escape_custom_object(self):
    con = self.connections[0]
    cur = con.cursor()

    class Foo(object):
        value = "bar"
        encoder = lambda x: x.value

    self.assertEqual(con.escape(Foo(), encoders={Foo: encoder}), "bar")
```

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -725,11 +725,11 @@ class Connection(object):
         self._execute_command(COMMAND.COM_INIT_DB, db)
         self._read_ok_packet()
 
-    def escape(self, obj):
+    def escape(self, obj, encoders=None):
         ''' Escape whatever value you pass to it  '''
         if isinstance(obj, str_type):
             return "'" + self.escape_string(obj) + "'"
-        return escape_item(obj, self.charset)
+        return escape_item(obj, self.charset, custom_encoders=encoders)
 
     def literal(self, obj):
         '''Alias for escape()'''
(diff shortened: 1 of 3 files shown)


methane:
Personally, I feel compatibility with MySQLdb is important only for APIs defined in DB-API 2.0, or
some ORM (SQLAlchemy, Django, Peewee or else) uses it.

It's OK to merge this API change.  But please copy not only API but also behavior.
Please look at https://github.com/PyMySQL/mysqlclient-python/blob/master/_mysql.c#L1169-L1186

```pycon
In [1]: import MySQLdb
In [2]: conn = MySQLdb.connect(user='root')
In [4]: import datetime
In [6]: m = {str: lambda x, d: "'" + str(x) + "'"}
In [7]: conn.escape(datetime.datetime.now(), m)
Out[7]: "'2015-02-05 23:30:49.041107'"
```

boglin:
Thanks; your approach seems sensible - I've updated the pull request to reflect this.

Some comments on the update:
 - It seems sensible that conn.encoders could be passed in as a mapping
 - 'escape_object' is now used for Decimal to provide the new interface for encoders
 - I've split the escape tests to a separate test case

'escape_set' from converters.py is not used since sets are handled by escape_sequence.  I left it in place, but wonder if it is still required?

PR #306, opened 2015-02-05
Multi-statement init_command causes subsequent query to fail

boglin:
Given that the CLIENT.MULTI_STATEMENTS flag is set, it is valid for the init_command supplied by the connection contain multiple statements.  Since on one ok_packet is read this causes subsequent queries to return unexpected results since client/server are out of sync.

Demonstrated by:
```
    def test_init_command(self):
        conn = pymysql.connect(
            init_command='SELECT "baz"; SELECT "boz"',
            **self.databases[0]
        )
        c = conn.cursor()
        c.execute('select "foobar";')
        self.assertEqual(('foobar',), c.fetchone())
        conn.close()
```

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -844,6 +844,9 @@ class Connection(object):
             if self.init_command is not None:
                 c = self.cursor()
                 c.execute(self.init_command)
+                # Wind pass any additional results
+                while self._result.has_next:
+                    self.next_result()
                 self.commit()
 
             if self.autocommit_mode is not None:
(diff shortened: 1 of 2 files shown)


pymysql/connections.py
```diff
@@ -844,6 +844,9 @@ def _connect(self):
             if self.init_command is not None:
                 c = self.cursor()
                 c.execute(self.init_command)
+                # Wind pass any additional results
+                while self._result.has_next:
+                    self.next_result()
```
methane: May `c.close()` solve the issue?
If it solves, I think it's clearer.

boglin: I think it would, yes.  I'll update the code.

methane:
Thanks

PR #307, opened 2015-02-06
[for comment] Performance improvements and reduction of client<->server roundtrips

boglin:
This pull request is focused on performance - particularly for an environment with high connection rates and, in some situations, high latency between database and client.

The changes include:
 - Refactoring the Connection class' `__init__`
 - Work on parsing the server information
 - Accumulation of variables which require setting on establishment of connection
 - Capability to defer any setup sql and prepend it to the user's first query.

Practically the deferred mode seems to give reasonable performance improvements - even to a locally hosted database.

Apologies for a fairly large change set - I've tried to prepare the four commits which make up the change to be reasonably logical steps through the implementation of the change.  The implementation is pretty complete - but there are a few rough edges/missing tests.  

At this point I'd like to get some opinions on the approach and what you'd like to see to have it accepted by the project.

Thanks for taking the time to look,
Dave

### DIFF
--- a/pymysql/_connection_protocol.py
+++ b/pymysql/_connection_protocol.py
@@ -1,0 +1,73 @@
+# Python implementation of the MySQL client-server protocol
+# http://dev.mysql.com/doc/internals/en/client-server-protocol.html
+
+import collections as _collections
+import struct as _struct
+
+
+def _parse_proto_desc(desc, clsName):
+    struct_types, fields = zip(*desc)
+    expr = "<" + "".join(struct_types)
+    non_empty_fields = (field for field in fields if field is not None)
+    proto_fields = " ".join(non_empty_fields)
+
+    newType = _collections.namedtuple(clsName, proto_fields)
+
+    return expr, newType
+
+
+def _expand_expression(expr, buffer):
+    while True:
+        pos = expr.find('z')
+        if pos < 0:
+            break
+        asciiz_start = _struct.calcsize(expr[:pos])
+        asciiz_len = buffer[asciiz_start:].find('\0')
+        expr = '%s%dsx%s' % (expr[:pos], asciiz_len, expr[pos+1:])
+    return expr
+
+
+def _parse(definition, ret_type, data):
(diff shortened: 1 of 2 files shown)


methane:
First of all, I have plan to use Cython to improve performance in 0.7.
Additionally, I'll add async APIs to connection in 0.7, too.

So I want to keep current code simple.
I'll check your changes on each commit and cherrypick if it is good and simple.

boglin:
Thanks - let me know if I can do anything my end to help.

s7v7nislands:
@methane After rewrite using cython, is it still support pypy?

methane:
@s7v7nislands Sure!
After having Cython speedup module, pure Python implementation can be more tuned for PyPy.
Additionally, I'll keep gevent compatibility.

methane:
@boglin 
* 383fea5 -- I like it.
* 5395b77 -- I feel `_connection_protocol` is over engineered for only two packet. plain `struct.unpack` is simple and readable enough.
* 3fcbd1d -- It's OK. `init_cmd` may be also combined to single query.
* 879f69f -- Interesting, but I don't want merge it since it makes hard to maintain PyMySQL.

When not using autocommit mode, BEGIN should not be executed.

When using autocommit mode, most query is executed in single query transaction.
BEGIN is called only when there are two or more queries executed in one transaction.
So overhead of BEGIN is limited.

So most important part of defer_mode is combining initial setup query and first query.
But I think it's not so important to accept this complexity.

Additionally, it can be achieved on higher level than PyMySQL.

```python
OWN_INIT_CMD = 'SET AUTOCOMMIT=1; SET sql_mode="STRICT_TRANS_TABLES";'
NUM_INIT_CMD = 2

# ...
conn = pymysql.connect(..., autocommit=None, ...)
with conn.cursor() as cur:
    cur.execute(OWN_INIT_CMD + sql, (args,))
    for _ in range(NUM_INIT_CMD):
        cur.nextset()
    res = cur.fetchall()
```

methane:
closed since no reply for long time.
Feel free to create new PR

PR #313, opened 2015-02-28
Need to use surrogate escape here too I think

wiggzz:
If bytes get passed in they are escaped using escape_bytes with a `surrogateescape` but then are not encoded properly here without this change.  This seems to fix the issue for me.


methane:
Wow, I'm sorry about the regression.
I hope I have time to release next version fixing this.

PR #328, opened 2015-03-20
Allow port to be a string

benjamin0:
This allows the port argument in the Connection constructor to be a string. For example:

```
Connection(port="3306")
```

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -606,7 +606,7 @@ class Connection(object):
             charset = _config("default-character-set", charset)
 
         self.host = host
-        self.port = port
+        self.port = int(port)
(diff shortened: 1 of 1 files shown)


methane:
Why?
I love Python since it's strict for type mismatch.
You can convert it explicitly.

```python
port = int(conf.get('port', 3306))
...
conn = pymysql.connect(..., port=port, ...)
```

benjamin0:
That's true, but it makes it a little easier if it's done automatically.

methane:
https://www.python.org/dev/peps/pep-0020/
>     Errors should never pass silently.
>     In the face of ambiguity, refuse the temptation to guess.

PR #332, opened 2015-04-01
Set the 'charset' in the example for starting a connection

MartinThoma:
I think this change is important, as it might prevent some problems users could run into.


methane:
Could you use utf8mb4?

PR #333, opened 2015-04-03
Don't try to iterate over none for warnings

mikemill:
When executing multiple statements that generate warnings
`show_warnings()` may produce `None` which `_show_warnings()` was trying to
iterate over.


graingert:
Can we get the warnings from the other statements?

methane:
@mikemill Thank you. I've merged this with some modifications.

PR #344, opened 2015-06-01
Enable no_delay by default and deprecate it

### LINKED ISSUE #334: Always use TCP_NODELAY and deprecate no_delay option
PyMySQL build packet on memory and send it in one system call.
Not using TCP_NODELAY doesn't reduce packet.
So I think we should use TCP_NODELAY always.

methane:
fixes #334


cactus:
:+1:

PR #351, opened 2015-06-27
expose get_server_version to return version tuple

grooverdan:
Seemed like such a good use of a tuple in the text may as well expose it.

I haven't looked where the documentation is generated from.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -15,6 +15,7 @@ import struct
 import sys
 import traceback
 import warnings
+import re
 
 from .charset import MBLENGTH, charset_by_name, charset_by_id
 from .cursors import Cursor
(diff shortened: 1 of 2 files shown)


methane:
For testing, I know which RDBS I use.  But I don't know all servers which uses MySQL protocol in the world.
If some server returns string doesn't match the regex, PyMySQL can't connect to the server.

I don't want to add public API without real use case.
Isn't `SHOW VARIABLES LIKE "version";` enough for you?

grooverdan:
I was mainly going to use this for internal uses - like #352.

See your point about not matching.

The tuple comparison is a good way for people to hard code comparisons against the major version number.

I've got one patch coming soon that has the following to avoid an implementation bug.
<pre>
            server_version_tuple > (5,5,10) and server_version_tuple > (5,6,2):
</pre>

Would you some decent error handling on the regex make this suitable?

grooverdan:
turns out i didn't need it

PR #352, opened 2015-06-27
eventually mysql will move beyond version 5

grooverdan:
Lets make sure that CLIENT.MULTI_RESULTS still works if mysql increases a version number


grooverdan:
replaced with #354

Issue #243, opened 2014-05-23
"Client does not support authentication protocol requested by server; consider upgrading MySQL client" when attempting to authenticate over sockets


oconnore:
Using: http://dev.mysql.com/doc/mysql-security-excerpt/5.5/en/socket-authentication-plugin.html

grooverdan:
fixed in #353

nicklasb:
The PR has not been merged yet, so this might be a bit premature.

grooverdan:
its merged now.

nicklasb:
Good stuff!

PR #353, opened 2015-06-28
Add authentication plugin support

grooverdan:
This extracts a plugin name from the server and exposes this as
connection.get_plugin_name().

The client side here adds capabilities PLUGIN_AUTH and
PLUGIN_AUTH_LENENC_CLIENT_DATA.

Because of this the HandshakeResponse response has been altered to use
the server_capabilities a bit more strictly.

The immediate upshot is that plugins like unix_socket are immediately
supported where they previously would of returned and error as without
PLUGIN_AUTH being advertised on the client side, the server only accepts
mysql_native_password and mysql_old_password.

To support other plugins some more work is needed to implement these.

I didn't intentionally break 4.X support  however I haven't tested it either.


grooverdan:
are you avoiding this one? Is there some reassurance I can help with?

pymysql/util.py
```diff
@@ -17,3 +17,18 @@ def join_bytes(bs):
         for b in bs[1:]:
             rv += b
         return rv
+
+# https://dev.mysql.com/doc/internals/en/integer.html#packet-Protocol::LengthEncodedInteger
+def lenenc_int(i):
```
methane: `util` module shouldn't know about MySQL protocol.
Please move this function to `connections` module.

pymysql/tests/test_connection.py
```diff
@@ -74,6 +76,102 @@ def test_init_command(self):
...
+                cur.execute("install plugin auth_socket soname 'auth_socket.so'")
+                socket_plugin_name = 'auth_socket'
+                socket_added = True
+            except pymysql.err.InternalError:
+                cur.execute("install soname 'auth_socket'")
+                socket_plugin_name = 'unix_socket'
+                socket_added = True
+
+        current_db = self.databases[0]['db']
+        db = copy.copy(self.databases[0])
```
methane: `dict.copy()` may be enough. `copy` module is not required.

methane:
> Is there some reassurance I can help with the acceptance of this patch?

* This PR is large.
* I don't have time to test authentication plugins.

pymysql/connections.py
```diff
@@ -345,7 +353,7 @@ def is_eof_packet(self):
         # http://dev.mysql.com/doc/internals/en/generic-response-packets.html#packet-EOF_Packet
         # Caution: \xFE may be LengthEncodedInteger.
         # If \xFE is LengthEncodedInteger header, 8bytes followed.
-        return len(self._data) < 9 and self._data[0:1] == b'\xfe'
+        return self._data[0:1] == b'\xfe'
```
methane: Why? Have you read comment above?

methane:
PyMySQL doesn't have enough test suite about authentication.
So I'm afraid of merging such large PR.

pymysql/connections.py
```diff
@@ -1014,31 +1051,105 @@ def _request_authentication(self):
...
+              return handler.authenticate(auth_pkt)
+            except AttributeError:
+                if plugin_name != 'dialog':
+                    raise err.OperationalError(2059, "Authentication plugin '%s'" +
+                              " not loaded: - missing authenticate method" % plugin)
+        else:
+            handler = None
+        if plugin_name == "mysql_native_password":
+            # https://dev.mysql.com/doc/internals/en/secure-password-authentication.html#packet-Authentication::Native41
+            data = _scramble(self.password.encode('latin1'), auth_packet.read_all()) + int2byte(0)
```
methane: `int2byte(0)` should be `b'\0'`

pymysql/connections.py
grooverdan: When the auth switch is done the length isn't < 9. http://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::AuthSwitchRequest

I've found the <9 reference here: http://dev.mysql.com/doc/internals/en/packet-OK_Packet.html. I'll see what can be done to avoid changing this bit and support the AuthSwitchRequest.

pymysql/util.py
grooverdan: lenenc-int appears in other places outside connection like https://dev.mysql.com/doc/internals/en/com-query-response.html, https://dev.mysql.com/doc/internals/en/binary-protocol-resultset.html and https://dev.mysql.com/doc/internals/en/table-map-event.html

grooverdan:
> PyMySQL doesn't have enough test suite about authentication.

Ok. no problem. I can add some more. Might get some code coverage there to identify the bits that need testing a little better.

> I don't have time to test authentication plugins.

I've extended the test suite a bit more to cover it.

>So I'm afraid of merging such large PR.

ok. Hoping adding more tests is a sufficient way to test this.

grooverdan:
Despite appearances this is still passing Travis tests and builds https://travis-ci.org/grooverdan/PyMySQL/builds/77866544 with a significant portion of the code added being covered https://coveralls.io/builds/3435133/source?filename=pymysql%2Fconnections.py .

I'll merge the changes with #369.

Is a sufficient test coverage of code added a suitable reassurance for merging this PR?

Would you prefer this to be broken down into smaller PRs?

grooverdan:
I'm finished. Questions/queries or requests for fixing code welcome.

nicklasb:
Just lurking this...
In my understanding, it is currently not possible to authenticate over sockets, which seems lika really severe problem(at least it is for me), which this PR seems to address.
 
And while neither I am not *fond* of huge PR:s, they aren't atomic but consists of lots of commits, and therefore  things can be selectively rolled back. 

To put it short, please merge this.  :-)

grooverdan:
it got quite big quickly in the number of commits. Happy to rebase into a single (or a few basic increments) commits. All of the code adds rigour to the protocol and callouts to where authentication plugins are required. In the mean time @nicklasb , the first commit 2d1da09f0781fb95479a9e07c047ff8f7cb975fc of adding PLUGIN_AUTH to the CAPABILITIES is sufficient to get socket_auth working.

nicklasb:
Thanks for your response, @grooverdan . 
Well, sometime that happens, especially if one continues to improve it. 

I am not sure that the maintainers have time to tell you how to break it apart..but as a maintainer of other projects myself, and if I were you, I would try to first break out each or all bug fixes in a separate PR, as those could then be released in a minor version.  

Then add structural changes and new functionality in another PR.

As there is not visible develop branch for the new stuff, that is the way I'd do it.

I can't deploy non-pip-stuff, but I can at least try it.

nicklasb:
Sorry, didn't work with just pasting over the 0.6.7 code, and I don't have time to clone and go on git now, perhaps I can test it tomorrow instead.

nicklasb:
Hi @grooverdan, I have now tried with this branch, and it works. 
However, it says that the MariaDB issue(#243) is resolved by this, but as this isn't merged, that isn't strictly true, and should as such be reopended.

pymysql/util.py
methane: All of these are protocol related.
And all code relating to protocol is in `connections` module for now.

grooverdan:
thanks @methane

methane:
While I've merged this pull request, I couldn't get how `connection.next_packet` works.
Until I understand it and refactoring it for future maintenance, I won't release next version.

nicklasb:
It's OK. Cool kids use #master. :-)

grooverdan:
>  I couldn't get how connection.next_packet works.

Its the next sequence number in the packet protocol (https://dev.mysql.com/doc/internals/en/sequence-id.html).

It seems that pkt_nr is cast down to uchar during use (despite being a uint in the code) - most of its use in this file: https://github.com/mysql/mysql-server/blob/5.7/sql/net_serv.cc#L789

methane:
Thanks for reply.
But I have researched it already and refactoring it now.
See #405.

PR #354, opened 2015-06-28
prepare for when mysql might want to move beyond version 5

grooverdan:
simpler version than #352


pymysql/connections.py
```diff
@@ -986,7 +986,7 @@ def _execute_command(self, command, sql):
 
     def _request_authentication(self):
         self.client_flag |= CLIENT.CAPABILITIES
-        if self.server_version.startswith('5'):
+        if int(self.server_version.split('.',1)[0]) >= 5:
```
methane: space should be after comma.

PR #355, opened 2015-06-28
Allow socket like options to be used

grooverdan:
Adds defer_connect (default False) arguement to connection constructor.

Move _connect method to public connect and add sock parameter.

The goal of this is to allow a mysql connection over an existing socket like object.

I'm a backer of the https://www.kickstarter.com/projects/1839007321/mysql-cli project that will provide an alternate mysql client. One of the feature requests is to run a mycli locally and get it to connect to the server over ssh. I was looking into https://github.com/paramiko/paramiko/blob/master/paramiko/channel.py for this. This Channel class presents all the same interfaces as a socket and passing this to PyMySQL will allow a direct connection to the remote server over ssh without an intermediate proxy locally.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -510,7 +510,7 @@ class Connection(object):
                  connect_timeout=None, ssl=None, read_default_group=None,
                  compress=None, named_pipe=None, no_delay=None,
                  autocommit=False, db=None, passwd=None, local_infile=False,
-                 max_allowed_packet=16*1024*1024):
+                 max_allowed_packet=16*1024*1024, defer_connect=False):
         """
         Establish a connection to the MySQL database. Accepts several
         arguments:
@@ -545,6 +545,8 @@ class Connection(object):
         autocommit: Autocommit mode. None means use server default. (default: False)
         local_infile: Boolean to enable the use of LOAD DATA LOCAL command. (default: False)
         max_allowed_packet: Max size of packet sent to server in bytes. (default: 16MB)
+        defer_connect: Don't explicitly connect on contruction - wait for connect call.
+            (default: False)
 
         db: Alias for database. (for compatibility to MySQLdb)
         passwd: Alias for password. (for compatibility to MySQLdb)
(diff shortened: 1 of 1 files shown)


pymysql/connections.py
```diff
@@ -818,31 +821,31 @@ def set_charset(self, charset):
...
-                        if e.errno == errno.EINTR:
-                            continue
-                        raise
-                self.host_info = "socket %s:%d" % (self.host, self.port)
-                if DEBUG: print('connected using socket')
-                if self.no_delay:
-                    sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
-            sock.settimeout(None)
-            sock.setsockopt(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)
+            if sock == None:
```
methane: `sock is None`

methane:
Thanks

grooverdan:
thank you  for merging it :-) if you require more info on other PRs just ask.

grooverdan:
Just wanted to say thanks, this now works (after paramiko/paramiko/pull/553 is added). really nice mysql over ssh tunnel which was easy.
<pre>
import paramiko
import pymysql

client = paramiko.SSHClient()
client.load_system_host_keys()
client.connect('XXX.XXX.com.au',username='danblack')
transport = client.get_transport()
channel = transport.open_channel('direct-tcpip',('127.0.0.1', 3306),('localhost', 4144))
c=pymysql.connect(database='information_schema', user='danblack', defer_connect=True)
c.connect(channel)
with c.cursor() as cursor:
    cursor.execute('SELECT VERSION()')
    print cursor.fetchone()
</pre>

gled-rs:
Connection launched
Now using the channel
  File "/usr/lib64/python3.3/site-packages/PyMySQL-0.6.6-py3.3.egg/pymysql/connections.py", line 924, in _write_bytes
    self.socket.sendall(data)
  File "/usr/lib64/python3.3/site-packages/paramiko-1.16.0-py3.3.egg/paramiko/channel.py", line 798, in sendall
  File "/usr/lib64/python3.3/site-packages/paramiko-1.16.0-py3.3.egg/paramiko/channel.py", line 752, in send
  File "/usr/lib64/python3.3/site-packages/paramiko-1.16.0-py3.3.egg/paramiko/channel.py", line 1112, in _send
OSError: Socket is closed

Trying to use an ssh tunnel with the above code and an internal ip address instead of 127.0.0.1

PR #358, opened 2015-07-23
[WIP] Async based on asyncore

grooverdan:
I saw #296 and felt guilty for writing patches that totally conflict with it.

I'm not sure if this is a 100% right but its a start based on the asyncore.dispatcher sockets.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -68,15 +68,14 @@ elif _py_version == (2, 6):
 
         def read(self, n):
             read = self._sock.recv(n)
-            if len(read) == n:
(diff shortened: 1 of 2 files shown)


methane:
Now I don't think supporting async operation in PyMySQL is good.
It may introduce significant performance penalty or large complexity to PyMySQL.

asyncio is new standard for async programming nowdays and there is [aiomysql](https://github.com/aio-libs/aiomysql) already.
Do you really need async support in PyMySQL?

grooverdan:
you're right. probably not needed.

PR #359, opened 2015-07-28
Add config parser to handle quotes in default file. Resolves #345

### LINKED ISSUE #339: pymysql.connect ignores host, connects to the machine's IP
```
>>> pymysql.connect(host='10.64.37.5', db="meta_p", read_default_file='~/.my.cnf')Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/nemobis/.local/lib/python3.4/site-packages/PyMySQL-0.6.6-py3.4.egg/pymysql/__init__.py", line 88, in Connect
    return Connection(*args, **kwargs)
  File "/home/nemobis/.local/lib/python3.4/site-packages/PyMySQL-0.6.6-py3.4.egg/pymysql/connections.py", line 644, in __init__
    self._connect()
  File "/home/nemobis/.local/lib/python3.4/site-packages/PyMySQL-0.6.6-py3.4.egg/pymysql/connections.py", line 838, in _connect
    self._request_authentication()
  File "/home/nemobis/.local/lib/python3.4/site-packages/PyMySQL-0.6.6-py3.4.egg/pymysql/connections.py", line 1021, in _request_authentication
    auth_packet = self._read_packet()
  File "/home/nemobis/.local/lib/python3.4/site-packages/PyMySQL-0.6.6-py3.4.egg/pymysql/connections.py", line 893, in _read_packet
    packet.check_error()
  File "/home/nemobis/.local/lib/python3.4/site-packages/PyMySQL-0.6.6-py3.4.egg/pymysql/connections.py", line 369, in check_error
    err.raise_mysql_exception(self._data)
  File "/home/nemobis/.local/lib/python3.4/site-packages/PyMySQL-0.6.6-py3.4.egg/pymysql/err.py", line 120, in raise_mysql_exception
    _check_mysql_exception(errinfo)
  File "/home/nemobis/.local/lib/python3.4/site-packages/PyMySQL-0.6.6-py3.4.egg/pymysql/err.py", line 112, in _check_mysql_exception
    raise errorclass(errno, errorvalue)
pymysql.err.OperationalError: (1045, "Access denied for user 'u2179'@'10.68.17.228' (using password: YES)")
```
I specified 10.64.37.5, but it tried to connect to 10.68.17.228 which is not the database server but my current machine. Originally I tried "s3.labsdb" which is the alias to the IP:
```
$ grep 10.64.37.5 /etc/hosts
10.64.37.5 s3.labsdb
```

https://github.com/PyMySQL/PyMySQL/issues/311#issuecomment-76492834 says the server needs to listen on a TCP socket, I don't know if it does.

Version:
```
$ easy_install3 --user pymysql
Searching for pymysql
Best match: PyMySQL 0.6.6
Processing PyMySQL-0.6.6-py3.4.egg
PyMySQL 0.6.6 is already the active version in easy-install.pth
```

### LINKED ISSUE #345: PyMySQL doesn't parse quoted config
Ref #339.

wraziens:


pymysql/__init__.py
```diff
@@ -31,6 +31,7 @@
 from .err import Warning, Error, InterfaceError, DataError, \
      DatabaseError, OperationalError, IntegrityError, InternalError, \
      NotSupportedError, ProgrammingError, MySQLError
+from .parsers import DefaultFileParser
```
methane: I don't want to export DefaultFileParser by this package.

pymysql/connections.py
```diff
@@ -17,11 +17,12 @@
 import warnings
 
 from .charset import MBLENGTH, charset_by_name, charset_by_id
-from .cursors import Cursor
 from .constants import CLIENT, COMMAND, FIELD_TYPE, SERVER_STATUS
-from .util import byte2int, int2byte
 from .converters import (
     escape_item, encoders, decoders, escape_string, through)
+from .cursors import Cursor
+from .parsers import DefaultFileParser
```
methane: `parsers` has too wide meanings.
Parsing option file, SQL and packet is completely different concern.
How about `pymysql.optionfile.Parser` ?

pymysql/parsers.py
```diff
@@ -0,0 +1,20 @@
...
+else:
+    import configparser
+
+
+class DefaultFileParser(configparser.RawConfigParser):
+
+    def __remove_quotes(self, value):
+        quotes = ["'", "\""]
+        for quote in quotes:
+            if value[0] == quote and value[-1] == quote:
```
methane: `if len(value) >= 2 and value[0] == value[-1] == quote`

pymysql/parsers.py
```diff
@@ -0,0 +1,20 @@
+from ._compat import PY2
+
+if PY2:
+    import ConfigParser as configparser
+else:
+    import configparser
+
+
+class DefaultFileParser(configparser.RawConfigParser):
+
```
methane: could you add this ?
```
def __init__(self):
    configparser.RawConfigParser.__init__(self, allow_no_value=True)
```

pymysql/parsers.py
```diff
@@ -0,0 +1,20 @@
...
+    def __remove_quotes(self, value):
+        quotes = ["'", "\""]
+        for quote in quotes:
+            if value[0] == quote and value[-1] == quote:
+                return value.strip(quote)
```
methane: How about `return value[1:-1].replace(r'\', '')` ?

pymysql/parsers.py
wraziens: ```allow_no_value``` was added in 2.7, does this mean 2.6 is no longer being supported?

methane: You're right.
I haven't dropped 2.6 support yet.
I think it should be happen in PyMySQL 0.7...

wraziens: If this will be part of the 0.7 release, I can add this in now. Or we can wait and add it later. What's your thoughts? Add it now or wait?

methane: Please wait for now.

wraziens: :+1:

methane:
Would you add simple test case?

pymysql/test/test_optionfile.py:

```python
from pymysql.optionfile import Parser
from unittest import TestCase

try:
    from cStringIO import StringIO
except ImportError:
    from io import StringIO


__all__ = ['TestParser']


_cfg_file =r"""
[default]
string = foo
quoted = "bar"
""")


class TestParser(TestCase):

    def test_string(self):
        parser = Parser()
        parser.readfp(StringIO(_cfg_file))
        self.assertEqual(parser.get("default", "string"), "foo")
        self.assertEqual(parser.get("default", "quoted"), "bar")
```

And add `from pymysql.tests.test_optionfile import *` to `pymysql/tests/__init__.py`.

wraziens:
Sure

methane:
Thanks

PR #361, opened 2015-07-31
Fix spelling in README example

Gnouc:
`pymysql.connect` function now use `password` as keyword argument instead of `passwd`.


methane:
Thanks

Issue #364, opened 2015-08-11
UnicodeDecodeError in executemany with mixed str/bytes arguments


hannosch:
There is workaround code for mixed str/bytes arguments in the execute (single) path, but this is missing from the executemany code.

PR with test to follow shortly.

grooverdan:
@hannosch 

I'm working on adding tests for mysql-5.6/5.7 (https://github.com/grooverdan/PyMySQL/tree/mysql56_57_tests) and get the following error with this test in py34 and mysql-5.7.

If you have any suggestions on fixing the test/implementation I'm keen to hear it

https://travis-ci.org/grooverdan/PyMySQL/jobs/77274309
<pre>
======================================================================
ERROR: test_issue_364 (pymysql.tests.test_issues.TestGitHubIssues)
Test mixed unicode/binary arguments in executemany.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/grooverdan/PyMySQL/pymysql/tests/test_issues.py", line 421, in test_issue_364
    cur.execute(sql, args=values)
  File "/home/travis/build/grooverdan/PyMySQL/pymysql/cursors.py", line 146, in execute
    result = self._query(query)
  File "/home/travis/build/grooverdan/PyMySQL/pymysql/cursors.py", line 297, in _query
    self._do_get_result()
  File "/home/travis/build/grooverdan/PyMySQL/pymysql/cursors.py", line 312, in _do_get_result
    self._show_warnings(conn)
  File "/home/travis/build/grooverdan/PyMySQL/pymysql/cursors.py", line 325, in _show_warnings
    warnings.warn(str(msg), err.Warning, 4)
pymysql.err.Warning: Invalid utf8 character string: 'FF00'
</pre>

hannosch:
@grooverdan I have no real idea. I've been able to reproduce the warning though. That is locally on OS X with MySQL 5.7 under Python 2.7.10. So it seems it's not just a Python 3 issue.

Ran out of time for now, will look into this more but don't know when I'll find the time.

grooverdan:
ok. No problem. for now I've caught the warning as expected for mysql-5.7

methane:
mysqlclient-python uses libmysqlclient.

```python
In [4]: import MySQLdb.connections
In [8]: conn = MySQLdb.connect(charset='utf8mb4')
In [13]: for i in range(0, 256, 16):
[...]

PR #321, opened 2015-03-04
Encoding error appears when iterable is passed in query args list.

zhiltsov-sigma:
To reproduce the issue put the list which contains two types of items: iterable containing unicode string and unicode string with non-ascii symbol

PR #365, opened 2015-08-11
Address encoding issues with mixed str/bytes in executemany and with iterables.

hannosch:
This should fix both #321 and #364.

I've incorporated the changes from #364 and hopefully addressed the review feedback. Since both of these issues touched the same code paths, one of them would have to be rebased on top of the other one.

Basically this pulls out the special PY2 handling out of the execute/mogrify function and also applies it to the executemany function. This all feels like a bit of a hack, but I don't know enough about the code to figure out how to better fix this. Pointers welcome :)

### DIFF
--- a/pymysql/cursors.py
+++ b/pymysql/cursors.py
@@ -95,9 +95,7 @@ class Cursor(object):
         return self._nextset(False)
 
     def _ensure_bytes(self, x, encoding=None):
-        if not PY2:
-            return x
-        if isinstance(x, unicode):
+        if isinstance(x, text_type):
             x = x.encode(encoding)
         elif isinstance(x, (tuple, list)):
             x = type(x)(self._ensure_bytes(v, encoding=encoding) for v in x)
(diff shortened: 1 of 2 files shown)


pymysql/cursors.py
```diff
@@ -93,14 +94,32 @@ def _nextset(self, unbuffered=False):
     def nextset(self):
         return self._nextset(False)
 
+    def _ensure_bytes(self, x, encoding=None):
+        if not PY2:
+            return x
```
methane: Is this `if not PY2` required?
This behavior is not predictable from method name.

hannosch: It's not required, as all the calling sites are protected by a `if PY2` check. I just put it there as the next line has `unicode` on it, which isn't even defined in Python 3, so it's more of a precaution if someone else where to refactor the code and call this from somewhere else. I'd be happy to remove the check here.

methane: text_type can be used to replace unicode.

methane:
Thanks

Issue #363, opened 2015-08-07
Binary types are attempting to be decoded into ascii causing a decoding exception.


danieljwest:
With use_unicode on and using a datatype that does not fall within the list of TEXT_TYPES the code is attempting to decode it as ascii. This causes some binary fields like geometry to cause an exception.

Attempting to select out of the below table will cause a decoding error 

CREATE TABLE table_xxx (
ID bigint(20) NOT NULL,
geom linestring NOT NULL,
PRIMARY KEY (ID),
SPATIAL KEY geom (geom)
) ENGINE=MyISAM DEFAULT CHARSET=utf8;

INSERT INTO table_xxx ( ID, geom) VALUES ( 1998, GeomFromText('LINESTRING(116.37604 39.73979,116.375 39.73965)'));

It's not clear to me whether or not the ascii encoding should be changed to None or if the binary check should be brought out of the string check.

If you let me know I can make a pull request for the change.

Here is the code in question from the connections.py file:
```python
            if use_unicode:
                if field_type in TEXT_TYPES:
                    charset = charset_by_id(field.charsetnr)
                    if charset.is_binary:
                        # TEXTs with charset=binary means BINARY types.
                        encoding = None
                    else:
                        encoding = charset.encoding
                else:
                    encoding = 'ascii'
            else:
                encoding = None
```

PR #368, opened 2015-08-15
Classify Geometry as a text type, refs #363.

hannosch:
I mixed up the issue number while working on 364, so corrected that test inside this PR.

Otherwise this replicates the example given in #363 and extends it to test for WKT/WKB values. Interestingly enough selecting for both WKT and WKB values worked before this change. It's just selecting for the internal MySQL binary geometry format that failed.

@danieljwest does this test and fix look good to you?

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -91,7 +91,8 @@ TEXT_TYPES = set([
     FIELD_TYPE.STRING,
     FIELD_TYPE.TINY_BLOB,
     FIELD_TYPE.VAR_STRING,
-    FIELD_TYPE.VARCHAR])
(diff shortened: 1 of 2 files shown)


danieljwest:
Gooood morning! 
As far as I can tell this looks good. I've ran it through with a number of binary fields and especially the geometry field and everything is returning as expected!

methane:
Thanks

danieljwest:
@methane Do you have any idea when this might make it into a release? I only ask because we have some software (mycli) that is breaking on this bug currently and we would like to get it fixed.

Thanks for your help!

methane:
released.

PR #369, opened 2015-08-26
Add MySQL-5.6 and 5.7 to test suite

grooverdan:
MySQL-5.7 has some warnings in a couple of cases so the test cases have been modified.

Geom functions give warnings now was ST_{function} is the newer syntax so just accept the warnings for now.

#364 I'm not sure if silencing those warnings is correct.

For other tests I've worked around the constraints of 5.7 by altering the test in a compatible way to 5.0+ and haven't interfered with the objective of the test.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,7 +2,8 @@
 branch = True
 source =
     pymysql
-
+omit = pymysql/test/*
+       pymysql/tests/thirdparty/test_MySQLdb/*
 
 [report]
 exclude_lines =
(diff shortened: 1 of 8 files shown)


grooverdan:
I'm pretty much ready for this to be merged.

It now covers a lot of the the code class (https://coveralls.io/github/grooverdan/PyMySQL?branch=mysql56_57_tests) and tests against multiple mysql versions.

I'm still working on SSL connection tests however I'll put them in a separate pull request. I'm getting all sorts of interesting errors on SSL connections (https://travis-ci.org/grooverdan/PyMySQL/builds/77890122).

methane:
Good job!
Thanks

Issue #371, opened 2015-08-26
Do not trim the string in result set


Ting-y:
What files should i look at to if i want to modify the return result set?

grooverdan:
connections.py and cursors.py.

Do you have a test case that shows string trimming?

Ting-y:
I have added a test case in my folk repository. What should i do next if i want to modify the result set

https://github.com/Ting-y/PyMySQL/commit/5aeea8fc5ea6f05913ea69c124a06c37e0d0cea4

methane:
I can't reproduce it...

```python
In [8]: cur.execute("CREATE TABLE test (data varchar(10))")
Out[8]: 0

In [9]: cur.execute("insert into test (data) values " "('string'), ('string   ')")
Out[9]: 2

In [10]: cur.execute("select data from test")
Out[10]: 2

In [11]: cur.fetchall()
Out[11]: (('string',), ('string   ',))

In [12]: cur.execute("select data from test")
Out[12]: 2

In [13]: cur.fetchone()
Out[13]: ('string',)

In [14]: cur.fetchone()
Out[14]: ('string   ',)
```

methane:
@Ting-y Could you send pull request for the test?
Travis-CI will run your test on some environments.

methane:
One MySQL pitfall is VARCHAR column ignores trailing spaces when comparison.

```python
In [15]: cur.execute("SELECT count(1) FROM test WHERE data='string'")
Out[15]: 1

In [16]: cur.fetchone()
Out[16]: (2,)
```

Ting-y:
Should I trim the space after fetch the result?

methane:
No.
Please send me a failing test.

2015年8月28日(金) 23:33 Ting Yang <notifications@github.com>:

PR #375, opened 2015-08-28
add test case to test string is not trim

Ting-y:
Test case for #371

### DIFF
--- a/pymysql/tests/test_issues.py
+++ b/pymysql/tests/test_issues.py
@@ -469,3 +469,26 @@ class TestGitHubIssues(base.PyMySQLTestCase):
         # don't assert the exact internal binary value, as it could
         # vary across implementations
         self.assertTrue(isinstance(row[0], bytes))
+    def test_issue_371(self):
+        """ Test return string is not trim. """
+        conn = self.connections[0]
+        self.safe_create_table(
+            conn,
+            "test", "create table test (data varchar(10))",
+        )
+        cursor = conn.cursor()
+        try:
+            cursor.execute(
+                "insert into test (data) values "
+                "('string'), ('string   ')")
+
+            cursor.execute("select data from test")
+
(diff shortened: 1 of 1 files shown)


methane:
Your test is passed on Travis.
PyMySQL doesn't trim strings.

Ting-y:
If i want to trim the strings, what is the best way to do it?

methane:
str.strip()

Ting-y:
It there any way that i can override or extend the fetch method? and should i modify in pymysql? Is there any reason why not trim the space in pymysql library?

methane:
I can't catch what you say.

Didn't #371 say "PyMySQL trims strings unexpectedly"?
Did you said "PyMySQL should trim strings (=breaks user data)"?

Ting-y:
Sorry for the confuse question.

I was trying to ask, how to make PyMySQL trim the strings and i want to modify the PyMySQL library instead change my own application.

methane:
Use CHAR instead of VARCHAR

Ting-y:
Its a good alternative, but i am not able to change my existing tables structure tho.

grooverdan:
@Ting-y , its PyMyQL's job to return data exactly how it is in the database. Modifying your own application or the data in the database is the correct solution. Some people expect trailing spaces to be returned so a change in behaviour isn't acceptable for PyMySQL.

If you'd like to clean up your data so no data columns contain leading or trailing spaces:
<pre>
Update tbl set data=TRIM(data)
</pre>

On application changes, If you'd like to return trimmed data you can use the str.strip() python method as mentioned earlier.

If you'd like to return a stripped data in a query:
<pre>
Select TRIM(data) from tbl
</pre>

Hope one of these methods satisfies your needs.

methane:
@Ting-y You can inherit cursor and wrap original fetch* methods.
You can pass your cursor class to connection through `cursorclass` keyword argument.

PR #377, opened 2015-09-02
Small fix in DBAPISet.__ne__

lzedl:


methane:
Why is this change required?

lzedl:
It's not correct to use `super` this way.
Correct with `super` (whithout `self`):
```Python
...
if isinstance(other, set):
    return super(DBAPISet, self).__ne__(other)
...
```
or like in pull request.

test code:
```Python
from pymysql import DBAPISet

a = DBAPISet([])
print a != set([])
```
now raises:
```
TypeError: expected 1 arguments, got 2
```

If you use super you don't need to pass `self` as first argument in method.

methane:
OK, Thanks.

PR #378, opened 2015-09-02
Add `column_names` property to Cursor class

drewbrew:
Implements the same property as mysql.connector's MySQLCursor class, which is just a shorthand for self.description.


methane:
I don't want to add public API only for compatibility with MySQL Connector/Python.
My order of priority is:

1. DB-API 2.0
2. compatibility with mysqlclient
3. Simplicity

PR #387, opened 2015-10-23
Annotate type and set self.connection to None in MySQLResult.read as …

tonal:
…other methods

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -1117,6 +1117,9 @@ class Connection(object):
 class MySQLResult(object):
 
     def __init__(self, connection):
+        """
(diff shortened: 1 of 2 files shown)


pymysql/connections.py
```diff
@@ -1117,6 +1117,9 @@ def get_server_info(self):
 class MySQLResult(object):
 
     def __init__(self, connection):
+        """
+        @type connection: Connection
```
methane: I prefer `:type connection: Connection` syntax.
https://www.jetbrains.com/pycharm/help/type-hinting-in-pycharm.html#d290117e152

PR #388, opened 2015-10-30
Add parameter for overriding encoders. Resolves #220

### LINKED ISSUE #220: encoders can't be overridden without mutating shared global state
`Connection.__init__` nicely accepts the `conv` argument which allows the *decoders* for a particular connection to be overridden.

There is no similar argument to the *encoders*.  Instead `pymysql.converters.escape_item` uses the global `encoders` dictionary.  The only way to change how values are encoded (for example, to change the datetime encoder to raise an exception whenever it is called with a naive datetime instance) is to mutate the global `pymysql.converters.encoders` dictionary.

It would be great if `Connection.__init__` also accepted an argument to override the encoders that are used so that they can be overridden on a per-connection basis.

wraziens:


methane:
This PR is incompatible with mysqlclient.
See [here](https://github.com/PyMySQL/mysqlclient-python/blob/master/MySQLdb/connections.py#L206-L207)

PR #391, opened 2015-11-14
TLS/SSL Fixes

klondi:
This pull merges a lot of SSL fixes:
* Allow TLSv1.1 or higher connections whilst not allowing SSLv2 and SSLv3 by default
* Allow passing custom SSLContexts in the ssl parameter
* Support for the cipher and capath parameter
* Allow enabling or disabling checking of the hostname


methane:
Thanks!
Could you add a space after comma?

klondi:
Sure thing!

klondi:
Here you go @methane :)

PR #395, opened 2015-11-24
Bump version_info

alazaro:
This will allow Django to use DateTimeField with microseconds.
It checks the driver version here:
https://github.com/django/django/blob/master/django/db/backends/mysql/features.py#L53


methane:
thanks

Issue #199, opened 2013-12-04
Add documentation


kevinburke:
I can't seem find the documentation for this library. There's an example.py file but no examples of the best way to do inserts/updates/deletes and check that they succeeded, how to catch common errors like invalid field type or duplicate key exception, how to escape values to avoid SQL injection, etc. There are links to DB-API but those don't explain the differences between DB-API and what was implemented here, for example DB-API doesn't mention the "passwd" field when connecting to databases.

ericdorsey:
Agreed; some documentation would be super nice -- for example, I didn't realize you needed to have ```autocommit=True``` as part of ```conn = pymysql.Connect()```, or a ```conn.commit()``` after a query to "save" after an insert. I found the answer [here on Stack Overflow](http://stackoverflow.com/questions/20485332/pymysql-insert-into-not-working/20485375#20485375), which looking at @mvader's response, is in https://www.python.org/dev/peps/pep-0249/ (the ```.commit()``` method is listed), but I don't see ```autocommit=True```, so not sure how the person from [this SO answer](http://stackoverflow.com/a/22680882/853178) knew that part? Is SO the de facto location of documentation for PyMySQL currently?

methane:
docstring

methane:
Contribution is welcome.
But I can't write good document since I'm not good English writer.

MartinThoma:
A documentation style should be chosen (and the choice should be documented).

I can see three common choices (by alphabetical order):

* Google Syntax ([example](http://sphinxcontrib-napoleon.readthedocs.org/en/latest/example_google.html))
  * Seems to be used by `natsorted`
* [Numpydoc syntax](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt)  ([example syntax](https://github.com/Lasagne/Lasagne/blob/master/lasagne/nonlinearities.py), [rendered example](http://lasagne.readthedocs.org/en/latest/modules/nonlinearities.html))
   * used by numpy, scipy
* Sphinx syntax
* No specific style:
  * logging
  * math
  * os

Besides personal preference, there are some criteria for the choice:

* Readability in the console when using `help`
* Support by automatic HTML documentation generation
* Support by editors
* How common the styles are / what people are used to

Anything else?

I couldn't find a guide how to use the google syntax / sphinx syntax and I don't know any project which consistantly used it.

Does anybody

See also:

[...]

PR #398, opened 2015-12-09
Add NumPyDoc docstring to BaseCursor.execute

MartinThoma:
* Parameters explained
* Return value explained

NumPyDoc is a standard docstring format (see [guide](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt)) which leads is very well readable in the source code, in the `help()`. It can also be used to generate beautiful documentation.

See the following examples  [`Lasagne/lasagne/layers/input.py`](https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/input.py) and [website docs](http://lasagne.readthedocs.org/en/latest/modules/layers.html)

### DIFF
--- a/pymysql/cursors.py
+++ b/pymysql/cursors.py
@@ -137,7 +137,25 @@ class Cursor(object):
         return query
 
     def execute(self, query, args=None):
-        '''Execute a query'''
+        """
+        Execute a query
+
+        Parameters
+        ----------
+        query : str
+            Query to execute on the server.
+        args : sequence or mapping, optional
+            Will be used with query
+
+        Returns
+        -------
+        int
+            Number of affected rows
+
+        Notes
+        -----
+        If args is a sequence, %s can be used as a placeholder in the query.
+        """
         while self.nextset():
             pass
 
--- a/pymysql/cursors.py
+++ b/pymysql/cursors.py
@@ -144,7 +144,7 @@ class Cursor(object):
         ----------
         query : str
             Query to execute on the server.
-        args : sequence or mapping, optional
+        args : tuple, list or dict, optional
             Will be used with query
 
         Returns
@@ -155,6 +155,7 @@ class Cursor(object):
         Notes
         -----
         If args is a sequence, %s can be used as a placeholder in the query.
+        If args is a dict, %(name)s can be used as a placeholder in the query.
         """
         while self.nextset():
             pass


MartinThoma:
I've just created this docstring (1) to give an example how it can be used (2) I wanted to know the return value of `cursor.execute` and it was not obvious.

MartinThoma:
Related to #199

methane:
I prefer sphinx's docstring format since sphinx autodoc supports it by default.

pymysql/cursors.py
```diff
@@ -137,7 +137,25 @@ def mogrify(self, query, args=None):
...
     def execute(self, query, args=None):
-        '''Execute a query'''
+        """
+        Execute a query
+
+        Parameters
+        ----------
+        query : str
+            Query to execute on the server.
+        args : sequence or mapping, optional
```
methane: `tuple, list or dict` since `_escape_args()` only supports them.

pymysql/cursors.py
```diff
@@ -137,7 +137,25 @@ def mogrify(self, query, args=None):
...
+            Will be used with query
+
+        Returns
+        -------
+        int
+            Number of affected rows
+
+        Notes
+        -----
+        If args is a sequence, %s can be used as a placeholder in the query.
```
methane: If args is a dict, %(name)s can be used.

MartinThoma:
@methane It is relatively easy to add numpydoc to sphinx (see https://codeandchaos.wordpress.com/2012/08/09/sphinx-and-numpydoc/) and numpydoc is MUCH easier to read.

Is there any official documentation generated by Sphinx by now? If not, I think it only matters how easier it is to read it in the command line with `help()`.

methane:
Why I prefer sphinx style is:

* Out of the box is better than plugin required.
* I know jedi and PyCharm supports sphinx style docstring.  I don't know how about NumPy syntax.

> Is there any official documentation generated by Sphinx by now?

mysqlclient-python has readthedocs site.  Currently, autodoc doesn't work but I want to support it in future.

methane:
There is also google style.
I like one good style than 10 better styles.

methane:
I researched quickly.

https://www.jetbrains.com/pycharm/help/python-integrated-tools.html
PyCharm 5 adds support for NumPy and Google style format. good.

http://jedi.jedidjah.ch/en/latest/docs/features.html#type-hinting
Jedi supports NumPy format, but it requires numpydoc plugin.
It means common installation of [jedi-vim](davidhalter/jedi-vim) doesn't support numpy syntax. -1.

MartinThoma:
@methane I like documentation at all better than no documentation. I've added a bit of documentation where nothing was before.

I think the discussion about the documentation style should be in #199 as it is a general question.

methane:
merged

PR #412, opened 2016-01-09
Update cursors.py

cherruq:
fix re that mismatch  INSERT  with ON DUPLICATE KEY


methane:
Please provide example which current PyMySQL can't handle.

Since PyMySQL doesn't have real parser, PyMySQL can't handle every cases.
It's a not bug, but a limitation.
Building query by yourself is better.

methane:
Close since no reply.
Please open again when you can explain what you want to do.

PR #420, opened 2016-01-16
added missing compat imports

blen2r:
The missing imports would cause crashes in our code base.


methane:
You shouldn't rely on them.
APIs not in `__all__` are private.

I suppose you should use [six](https://pypi.python.org/pypi/six).

PR #427, opened 2016-02-17
bug with parameters view as %(name)s for executemany

WorldException:
Only "%s" recognized with regular expression, but for "%(name)s" bulk insert will not work.
INSERT ..... VALUES (%(name)s, %(name)s)

I'm modify regular expression for recognize %s and %(name)s too.


methane:
Could you add test?

WorldException:
I added tests and check them.
I have unclean code in cursors.py.__init__()
self.max_stmt_length = connection.max_allowed_packet - 10000  # I don'n known packet_header_size

Maybe there a better way?

pymysql/cursors.py
```diff
@@ -40,6 +40,7 @@ def __init__(self, connection):
         self._executed = None
         self._result = None
         self._rows = None
+        self.max_stmt_length = connection.max_allowed_packet - 10000  # I don'n known packet_header_size
```
methane: Remove the comment.
packet header is 4 bytes.

WorldException:
I have one assert, may be it's another error?

Testing started at 11:45 ...

Process finished with exit code 0

Failure
  File "/home/quick/PycharmProjects/avtoformula/libs/PyMySQL/pymysql/tests/test_cursor.py", line 42, in test_cleanup_rows_unbuffered
    self.assertGreater(len(log), 0)
AssertionError: 0 not greater than 0

pymysql/connections.py
```diff
@@ -768,6 +768,19 @@ def show_warnings(self):
...
+        """Get settings from server"""
+        self._execute_command(COMMAND.COM_QUERY, "SHOW VARIABLES WHERE VARIABLE_NAME = %s" % self.escape(name))
+        result = MySQLResult(self)
+        result.read()
+        if result.rows:
+            variable = result.rows[0][1]
+            if variable.isdigit():
+                return int(variable)
+            else:
+                return variable
```
methane: Convert at caller site.

pymysql/connections.py
```diff
@@ -768,6 +768,19 @@ def show_warnings(self):
         result.read()
         return result.rows
 
+    def get_system_variable(self, name, default=None):
+        """Get settings from server"""
+        self._execute_command(COMMAND.COM_QUERY, "SHOW VARIABLES WHERE VARIABLE_NAME = %s" % self.escape(name))
```
methane: Use `"SELECT @@" + name` query

pymysql/cursors.py
```diff
@@ -158,6 +159,8 @@ def executemany(self, query, args):
 
         m = RE_INSERT_VALUES.match(query)
         if m:
+            if self.max_stmt_length == Cursor.max_stmt_length:
+                self.max_stmt_length = self.connection.get_system_variable("max_allowed_packet") - self.packet_header_size
```
methane: Stop using `self.max_stmt_length`.
Get max_allowed_packet from connection always.

methane: Using `max_allowed_packet` may eat too many memory.
We should have more reasonable soft limit.  (I think 1MB is reasonable).

methane:
@WorldException Could you split this PR into two?  I can merge `%(name)s` support soon.
But I want more time to think about using `max_allowed_packet` system var.

WorldException:
Can I make new branch with %(name)s changes only and create new pull request and cancel current pull request?
Or I need use another method?
I'm have not big expiriense with git for do this.

WorldException:
I have new idea about max_allowed_packet.
preview http://pastebin.com/3bEKr0n9
It's fix bug if server value  'max_allowed_packet' < self.max_allowed_packet and we can limit max_allowed_packet when create Connection().

methane:
I'm listless about using system variable automatically for now.

I think low level library must be simple, not be nosy.  And this MySQL Driver is the lowest level library.
Higher level wrapper library can do it for you.

Is there any strong reason to use server-side `max_allowed_packet`?
Even libmysqlclient doesn't do it.

PR #434, opened 2016-03-04
connections: Fixup parsing of incomplete rows

baloo:
Since mysql-5.6.2 (commit id: 6deae781c4c8e7018a31d56bc050f99e2411cbc0),
mysql may reply with incomplete row to "SHOW SLAVE HOSTS" commands.
This happens if slave_uuid does not exists, this happens if the slave
does not issue this command:
  set @slave_uuid='UUID'

See:
  https://github.com/mysql/mysql-server/commit/6deae781c4c8e7018a31d56bc050f99e2411cbc0#diff-a06ba9f19ff21c6cdb9bcfb44cc185cfR708

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -106,6 +106,9 @@ DEFAULT_CHARSET = 'latin1'
 
 MAX_PACKET_LEN = 2**24-1
 
+class _NoMoreColumnsInRow(Exception):
+    pass
+
 
 def dump_packet(data): # pragma: no cover
     def is_ascii(data):
@@ -332,7 +335,11 @@ class MysqlPacket(object):
         Length coded numbers can be anywhere from 1 to 9 bytes depending
         on the value of the first byte.
         """
-        c = self.read_uint8()
+        try:
+            c = self.read_uint8()
+        except IndexError:
+            raise _NoMoreColumnsInRow()
+
         if c == NULL_COLUMN:
             return None
         if c < UNSIGNED_CHAR_COLUMN:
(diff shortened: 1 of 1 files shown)


baloo:
I tried to implement a test for this, I couldn't use the actual SHOW SLAVE HOSTS command because in the current mysql travis configuration of PyMySQL there is no setting like log-bin. Which prevents me to actually connect a slave and have results in the command.
I couldn't find any other location in the mysql source code which would allow me to reproduce this. All responses looks like having the correct number of columns but this specific one.

pymysql/connections.py
```diff
@@ -332,7 +335,11 @@ def read_length_encoded_integer(self):
         Length coded numbers can be anywhere from 1 to 9 bytes depending
         on the value of the first byte.
         """
-        c = self.read_uint8()
+        try:
+            c = self.read_uint8()
+        except IndexError:
+            raise _NoMoreColumnsInRow()
+
```
methane: This is very performance critical part.
I worry about this `try`~`except` decrease PyMySQL performance lot.
Could you benchmark about SELECT many rows?

baloo: I understand your concerns, still `try` ~ `except` is only a performance issue if an exception is raised. As far as I can read the mysql source code, this is only triggered by the `SHOW SLAVE HOSTS` command if a pre-5.6 slave is connected behind a >=5.6.2 master (thus not submitting `set @slave_uuid='UUID'` command before `COM_BINLOG_DUMP`).
I believe this is an unfairly uncommon command and the case described is very rare, but I understand your concerns and I will run benchmarks. Do you have an automated setup / reporting test? or just timing select on 1M rows is fine with you?

methane:
I doubt that this is a bug of MySQL?
The commit you pointed out is too long and hard to me to check what happen in the commit.

Does current MySQL 5.7 have same issue?

baloo:
The commit I pointed out introduced the line you may find here:
https://github.com/mysql/mysql-server/blob/6deae781c4c8e7018a31d56bc050f99e2411cbc0/sql/repl_failsafe.cc#L708-709

A month later, the commit https://github.com/mysql/mysql-server/commit/8f629bd2b39f5ffea58f5d7c3b8b6460b8b85cd7 has refactored this code to what is currently in mysql5.7

Mysql5.7 still have the issue: 
https://github.com/mysql/mysql-server/blob/5.7/sql/rpl_master.cc#L251-L252

methane:
Quick benchmark.
I tried this simple bench to the `world` database[1].

[1]: https://dev.mysql.com/doc/index-other.html

There are no notable difference on PyPy 5.1.
In case of CPython 3.5.1, difference is 8.0~1sec vs 8.3~8.4sec.

PR #441, opened 2016-04-04
bit conversion converts to int not string

eatonphil:
I have some legacy code to support that uses a bit field as a boolean. I am switching from oursql (which automatically converts bit fields to integers) to pymysql. The field is registered in SqlAlchemy as a Boolean field. It gets properly put into the database but when it comes out there is a type error saying `an integer is required (got bytes)`. Tests failing in SqlAlchemy drove me nuts until I realized pymysql simply does nothing to convert bit fields from strings to python values. Oursql converts bit fields to integers because it [recognizes](https://github.com/python-oursql/oursql/blob/master/oursqlx/statement.pyx#L537) there is probably no other use for it. Since no solution has been proposed other than to ignore the issue, I propose that we start with this one.

Thank you!

### DIFF
--- a/pymysql/converters.py
+++ b/pymysql/converters.py
@@ -309,13 +309,8 @@ def through(x):
     return x
 
 
-#def convert_bit(b):
-#    b = "\x00" * (8 - len(b)) + b # pad w/ zeroes
-#    return struct.unpack(">Q", b)[0]
-#    
-#     the snippet above is right, but MySQLdb doesn't process bits,
-#     so we shouldn't either
-convert_bit = through
+def convert_bit(b):
+    return ord(b)
 
 
 def convert_characters(connection, field, data):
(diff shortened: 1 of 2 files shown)


pymysql/converters.py
```diff
@@ -309,13 +309,8 @@ def through(x):
     return x
 
 
-#def convert_bit(b):
-#    b = "\x00" * (8 - len(b)) + b # pad w/ zeroes
-#    return struct.unpack(">Q", b)[0]
-#    
-#     the snippet above is right, but MySQLdb doesn't process bits,
-#     so we shouldn't either
```
methane: Have you read this comment?

eatonphil: So is oursql wrong? Am I depending on bad behavior?

methane: No.  There is no "right" or "wrong" about this.
DB API 2.0 doesn't specify about how to convert bit type.
So this is up to driver's decision.

In case of PyMySQL, I follow MySQLdb's behavior as possible.

methane: MySQL-python (aka, MySQLdb) author is inactive for these years.
I maintains fork of it named mysqlclient-python.
But I won't break backward compatibility until I am official maintainer of MySQLdb.

eatonphil: Is my only option to maintain a fork or is there a better way to deal with this?

methane: 1) You can manually convert result value.
2) You can override converter mapping.

But I don't recommend (2).  Converter function can't know connection encoding and column encoding.  I think the API is broken.

Issue #27, opened 2011-07-27
Documentation


petehunt:
Originally filed by floydoph...@gmail.com on 2010-07-26T19:51:37

PyMySQL really should have inline docstrings and comments.

PR #458, opened 2016-05-17
Start using readthedocs

MartinThoma:
See #27 and #199.

The Sphinx documentation is only a draft. If you like it, I can adjust it (e.g. add `converters`, `cursors`, `cursors`, `util`) and probably move the examples from the README to the Sphinx docs.


methane:
@MartinThoma FWIW, do you know `--force-with-rease` option of `git pull` command?
No reason to create new pull request after rebase.

docs/requirements.txt
```diff
@@ -0,0 +1 @@
+pymysql
```
methane: Why do you install pymysql from PyPI instead of using current repository?

MartinThoma:
@methane No, I didn't know that. Thank you.

methane:
See http://docs.readthedocs.io/en/latest/faq.html#can-i-document-a-python-package-that-is-not-at-the-root-of-my-repository

methane:
@MartinThoma mention me when review ready.

methane:
@MartinThoma Did you check admin page of readthedocs?
"Install your project inside a virtualenv using setup.py install" may help you.

docs/requirements.txt
```diff
@@ -0,0 +1 @@
+git+git://github.com/PyMySQL/PyMySQL.git
```
methane: `..` is enough, maybe.

methane: Or use "Install your project inside a virtualenv using setup.py install".

MartinThoma:
@methane The documentation draft is ready so that one can get a feeling for how it might look like when everything would be done this way. If you are interested in getting this documentation, then (and only then) I will do so for the other modules. Are you interested?

methane:
@MartinThoma Having RTD is nice.  But don't use pymysql.readthedocs.io for preview.
Unless you add me to the project admin of RTD, I can't merge this pull request and I can't use pymysql.readthedocs.io domain.

methane:
BTW, I already have readthedocs for mysqlclient
https://mysqlclient.readthedocs.io/en/latest/#

There are no need of `docs/requirements.txt`.  Please see my comment above.

methane:
Don't strip information from README!!
README should provide quick and **complete** information.
Duplication between README and RTD is not a problem.

MartinThoma:
@methane I've added you as an admin on readthedocs.

> Don't strip information from README!!
README should provide quick and complete information.
Duplication between README and RTD is not a problem.

Are you sure you want it like this? Because then you might have to change always things in two places and it also makes the README longer.

methane:
> I've added you as an admin on readthedocs.

Thanks.

> Are you sure you want it like this? Because then you might have to change always things in two places and it also makes the README longer.

Yes.  Despite having doc in another site, README should provides enough information for people
choosing library they use.

Information for developer can be removed from README.  It includes:

* Install from git checkout
* Running test suite
* Link for MySQL protocol

Other information (Requirements, install from PyPI, Link to DB-API, MySQL, ML and license) should be in README.

methane:
Thanks.  I've merged and organized the docs.

MartinThoma:
Nice :-) I'm happy I could help.

PR #428, opened 2016-02-18
Added JSON field support

desertkun:
As of MySQL 5.7.8, MySQL supports a native [JSON](https://dev.mysql.com/doc/refman/5.7/en/json.html) field type, I've added support for this field. It gets automatically unpacked using `json.loads`, and falls back to `{}` in case of corrupted JSON.

According to [this](https://github.com/mysql/mysql-server/blob/a2757a60a7527407d08115e44e889a25f22c96c6/libbinlogevents/export/binary_log_types.h#L51) the `JSON` field type has code `MYSQL_TYPE_JSON=245`.

Thank you!

PR #477, opened 2016-06-24
Add JSON field type and include it in TEXT_TYPES list to prevent encoding problems

bmamouri:
**Note:** This issue is not related to the following pull request: https://github.com/PyMySQL/PyMySQL/pull/428/commits

MySQL 5.7 added JSON types. In `_get_descriptions` method we are setting the encoding of any field type that is not textual to 'ascii'. This is clearly wrong as the JSON field types are actually BINARY column that contains strings.

Because of wrong encoding problem, in `_read_row_from_packet` function we encode the JSON field type to 'ascii'. Here is the related code:

```
    def _read_row_from_packet(self, packet):
        row = []
        for encoding, converter in self.converters:
            try:
                data = packet.read_length_coded_string()
            except IndexError:
                # No more columns in this row
                # See https://github.com/PyMySQL/PyMySQL/pull/434
                break
            if data is not None:
                if encoding is not None:  # Since encoding is set 'ascii' the content is encoding to 'ascii'
                    data = data.decode(encoding)
                if DEBUG: print("DEBUG: DATA = ", data)
                if converter is not None:
                    data = converter(data)
            row.append(data)
        return tuple(row)
```

Because of this problem, querying JSON field types that contain Unicode characters will fail.

The solution to the problem is relatively easy. Add the JSON field types to TEXT_TYPES.


methane:
Thanks!

methane:
I'm sorry, but I revert this.

You should use `encoding="utf8mb4"` or `use_unicode=False` to avoid this issue.

PR #481, opened 2016-06-30
Trivial grammar change to cursors.py

krrg:


pymysql/cursors.py
```diff
@@ -454,7 +454,7 @@ def fetchall(self):
 
     def fetchall_unbuffered(self):
         """
-        Fetch all, implemented as a generator, which isn't to standard,
+        Fetch all, implemented as a generator, which isn't too standard,
```
methane: I can't understand what `too` means here.

I think this sentence means `fetchall_unbeffered()` is not defined in PEP 249.
(But I'm not sure because I'm not original author.)

krrg:
The original sentence doesn't quite make sense either, but your explanation makes sense. I was using "too" as in "Implemented as a generator, which is not a very standard".

PR #487, opened 2016-07-18
Avoid IEEE754 converting fraction of seconds

timonwong:
Due to the limitations of IEEE754, we may lost precision when
converting string back to datetime objects, for example:

`float('0.511581') * 1e6`, gives us `0.511580`, which is unexpected.

In order to address that issue, use of `float` is avoided.

### DIFF
--- a/pymysql/converters.py
+++ b/pymysql/converters.py
@@ -2,6 +2,7 @@ from ._compat import PY2, text_type, long_type, JYTHON, IRONPYTHON, unichr
 
 import datetime
 from decimal import Decimal
+import re
 import time
 
 from .constants import FIELD_TYPE, FLAG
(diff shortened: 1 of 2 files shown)


methane:
Thanks.  But I feel parsing code is large enough to use regular expression.
Could you rewrite to use this?

`r"(\d{2,4})-(\d{1,2})-(\d{1,2})[T ](\d{1,2}):(\d{1,2}):(\d{1,2})(?:.(\d{1,6}))?"`

timonwong:
> Thanks. But I feel parsing code is large enough to use regular expression.
> Could you rewrite to use this?
> 
> r"(\d{1,4})-(\d{1,2})-(\d{1,2})[T ](\d{1,2}):(\d{1,2}):(\d{1,2})(?:.(\d{1,6}))?"

Yep, I rewrited both timedelta and time converters as well.

methane:
Thank you.

PR #491, opened 2016-07-25
Warning propagation improvements

vtermanis:
Dear PyMySQL team,

Here are a couple of suggestions (similar to what I submitted to [MySQLdb#101](https://github.com/PyMySQL/mysqlclient-python/pull/101) regarding warning handling in your library:

1. _Include error code in warnings, not just string_  
I thought it would make sense for the `Warning` instances to not just have a string argument but, same as with errors, to also include the error code (as first argument, with string being second).
2. Fix warning propagation for unbuffered queries  
Since warning count is not available until whole result has been fetched, handle warnings "_deferred_" (simiarly to MySQldb)

### DIFF
--- a/pymysql/cursors.py
+++ b/pymysql/cursors.py
@@ -30,6 +30,8 @@ class Cursor(object):
     #: Default value of max_allowed_packet is 1048576.
     max_stmt_length = 1024000
 
+    _defer_warnings = False
+
     def __init__(self, connection):
         '''
         Do not create an instance of a Cursor yourself. Call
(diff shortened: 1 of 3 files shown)


vtermanis:
Apologies for failures - I shall investigate

vtermanis:
The problem was due to a warning not being propagated with an unbuffered cursor until the next query has been started (both DML, see change in `test_DictCursor.py`). I think this my change to move the next statement into the warning-catch clause is OK. (I think previously it didn't raise a warning at all for `SSDictCursor` in that particular case.)

I also added a unit test - hope that's the right place for it. Also, is the "coverage decreased" warning a red herring or have I done something wrong?

PR #497, opened 2016-08-09
 the demo missing import pymysql

wwj718:


methane:
Have you tried it?
`import foo.bar` implies `import foo`.
Because you can't access foo.bar without foo.

wwj718:
@methane I am sorry . It's my mistake

Issue #507, opened 2016-08-31
PyMySQL 0.7.7 is injecting strange codes into exception messages


zzzeek:
Exception messages have changed as of PyMySQL 0.7.7 such that some kind of codes are present right at the start of the message.   This breaks code that is looking for text within the message to distinguish the exact kind of error this is as well as parsing out the arguments.

Example, the following two errors in 0.7.6 look like this:

     (1054, u"Unknown column 'fake' in 'field list'")
    (1452, u'Cannot add or update a child row: a foreign key constraint fails (`test`.`b`, CONSTRAINT `b_ibfk_1` FOREIGN KEY (`a_id`) REFERENCES `a` (`id`))')

in 0.7.7, we get:

    (1054, u"42S22Unknown column 'fake' in 'field list'")
    (1452, u'23000Cannot add or update a child row: a foreign key constraint fails (`test`.`b`, CONSTRAINT `b_ibfk_1` FOREIGN KEY (`a_id`) REFERENCES `a` (`id`))')

those codes seem to be fixed so perhaps they are some kind of debugging string that forgot to be removed, but this is breaking a lot of downstream stuff.

test case:

```
import pymysql
import warnings

warnings.simplefilter("ignore")


conn = pymysql.connect(user="scott", passwd="tiger", host="127.0.0.1", db="test")

cursor = conn.cursor()


def show_exception(fn, *args):
    try:
        fn(*args)
    except Exception as err:
        print(err)

show_exception(
    cursor.execute, "select fake"
)


cursor.execute("""
    CREATE TABLE IF NOT EXISTS a (
       id INTEGER PRIMARY KEY
    ) ENGINE=InnoDB
""")

cursor.execute("""
    CREATE TABLE IF NOT EXISTS b (
       id INTEGER PRIMARY KEY,
       a_id INTEGER,
       FOREIGN KEY (a_id) REFERENCES a(id)
    ) ENGINE=InnoDB
""")

show_exception(
    cursor.execute, "INSERT INTO b (id, a_id) VALUES (1, 25)"
)
```

dianaclarke:
It's the removal of the following `is_41` code block in commit https://github.com/PyMySQL/PyMySQL/commit/02910b74bb4f776b0a71d0ab71d9a5b99273685e that's behind the exception message changes.

https://github.com/PyMySQL/PyMySQL/blob/c3ed30135117a13d5e52087d67249a252317c30a/pymysql/err.py#L100

[...]

PR #508, opened 2016-08-31
Revert error packet parsing changes

dianaclarke:
Pull request for issue #507.

Revert the changes made in commit 02910b7 to the mysql error packet
parsing, as it didn't take the CLIENT_PROTOCOL_41 case into account. 

It appears that the non-CLIENT_PROTOCOL_41 case was off by one as well
(sliced data on 4 vs 3).

I also added two new tests to cover those cases.

See: https://dev.mysql.com/doc/internals/en/packet-ERR_Packet.html

### DIFF
--- a/pymysql/err.py
+++ b/pymysql/err.py
@@ -97,6 +97,11 @@ del _map_error, ER
 
 def raise_mysql_exception(data):
     errno = struct.unpack('<h', data[1:3])[0]
-    errval = data[4:].decode('utf-8', 'replace')
+    is_41 = data[3:4] == b"#"
+    if is_41:
+        # client protocol 4.1
+        errval = data[9:].decode('utf-8', 'replace')
+    else:
+        errval = data[3:].decode('utf-8', 'replace')
     errorclass = error_map.get(errno, InternalError)
     raise errorclass(errno, errval)
--- a/pymysql/tests/test_err.py
+++ b/pymysql/tests/test_err.py
@@ -1,0 +1,21 @@
+import unittest2
+
+from pymysql import err
+
+
+__all__ = ["TestRaiseException"]
+
+
+class TestRaiseException(unittest2.TestCase):
+
+    def test_raise_mysql_exception(self):
+        data = b"\xff\x15\x04Access denied"
+        with self.assertRaises(err.OperationalError) as cm:
+            err.raise_mysql_exception(data)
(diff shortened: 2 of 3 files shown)


dianaclarke:
PS. working on the py26 and py3 cases now.

pymysql/tests/test_err.py
```diff
@@ -0,0 +1,21 @@
...
+from pymysql import err
+
+
+__all__ = ["TestRaiseException"]
+
+
+class TestRaiseException(TestCase):
+
+    def test_raise_mysql_exception(self):
+        data = "\xff\x15\x04Access denied"
```
methane: Use `b""` string.

dianaclarke: Yup ;) On it, silly me.

pymysql/err.py
```diff
@@ -97,6 +97,11 @@ def _map_error(exc, *errors):
 
 def raise_mysql_exception(data):
     errno = struct.unpack('<h', data[1:3])[0]
-    errval = data[4:].decode('utf-8', 'replace')
+    is_41 = data[3:4] == b"#"
+    if is_41:
+        # client protocol 4.1
+        errval = data[9:].decode("utf8", 'replace')
+    else:
+        errval = data[3:].decode("utf8", 'replace')
```
methane: Use utf-8 instead of utf8. (utf8 is alias).

```
In [1]: %timeit b"fizzbuzz".decode('utf-8', 'replace')
1000000 loops, best of 3: 341 ns per loop
In [2]: %timeit b"fizzbuzz".decode('utf8', 'replace')
1000000 loops, best of 3: 373 ns per loop
```

pymysql/tests/test_err.py
```diff
@@ -0,0 +1,21 @@
+from unittest import TestCase
```
methane: Use unittest2

pymysql/err.py
dianaclarke: Sure, I can do that quickly. I just grabbed the old code & didn't look all that hard at it ;)

pymysql/tests/test_err.py
dianaclarke: Yup, that was in my most recent push.

methane:
Thanks!

Issue #512, opened 2016-09-03
0.7.8 and 0.7.7 exception: pymysql.err.InternalError: Packet sequence number wrong - got 142 expected 1


Zeusw:
code：
```
import pymysql.cursors

connection = pymysql.connect(host='localhost',
                             user='root',
                             password='123456',
                             db='wikiurl',
                             charset='utf8mb4')

try:
    with connection.cursor() as cursor:
        sql = "select urlname, urlhref from urls where id is not null"
        conut = cursor.execute(sql)
        print(conut)
finally:
    connection.close()
```

This code in 0.7.6 is OK, but in 0.7.7 or 0.7.8 is not ok.
I think is, `id is not null ` in 0.7.7 is not supported.

methane:
It seems duplicate of #511, but this is more easy to reproduce.
Thanks!

methane:
Hmm, I can't reproduce.
"id is not null" can be used in PyMySQL 0.7.8.

mysqldump is required to reproduce.

Zeusw:
thank you for your reply.
I found the problem.
`
INSERT INTO `wikiurl`.`urls2` (`id`, `urlname`, `urlhref`) VALUES ('137', '', 'https://en.wikipedia.org/wiki/Main_Page');
`
if urlname is '',There will be an exception:pymysql.err.InternalError: Packet sequence number wrong - got 6 expected 1

Note: Is Null character, not NULL.

methane:
Could you show `show create table wikiurl`?

Zeusw:
id  type:int  length:11 Primary Key
urlname type:varchar length:255
urlhref type:varchar length:1000

Zeusw:
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                                                            |
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| urls  | CREATE TABLE `urls` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `urlname` varchar(255) DEFAULT NULL,
  `urlhref` varchar(1000) DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=306 DEFAULT CHARSET=utf8mb4 |
+-------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+

[...]

PR #513, opened 2016-09-03
Revert DEPRECATE_EOF support

methane:
0.7.7 introduce DEPRECATE_EOF support.  But it's broken.
It breaks reading rows which first column is empty string.

fixes: #511, #512


Zeusw:
Thank you for repair

PR #529, opened 2016-11-15
Allow connecting to the DB from a specific network interface

dciabrin:
When connecting via TCP/IP, provide a means to the client to bind the
socket to be created to a specific source address.  This is useful
when the client host has multiple network interfaces and a specific
one must be used for database traffic.

Add a new optional "bind" argument to class Connection, to specify the
interface from which to connect to the DB. Argument can be a hostname
or an IP address.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -534,7 +534,8 @@ class Connection(object):
                  compress=None, named_pipe=None, no_delay=None,
                  autocommit=False, db=None, passwd=None, local_infile=False,
                  max_allowed_packet=16*1024*1024, defer_connect=False,
-                 auth_plugin_map={}, read_timeout=None, write_timeout=None):
+                 auth_plugin_map={}, read_timeout=None, write_timeout=None,
+                 bind=None):
         """
         Establish a connection to the MySQL database. Accepts several
         arguments:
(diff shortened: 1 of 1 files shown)


zzzeek:
just a note we are imitating MySQL's own API here, which should also be supported in mysqlclient:  MYSQL_OPT_BIND at dev.mysql.com/doc/refman/5.7/en/mysql-options.html

pymysql/connections.py
```diff
@@ -534,7 +534,8 @@ def __init__(self, host=None, user=None, password="",
                  compress=None, named_pipe=None, no_delay=None,
                  autocommit=False, db=None, passwd=None, local_infile=False,
                  max_allowed_packet=16*1024*1024, defer_connect=False,
-                 auth_plugin_map={}, read_timeout=None, write_timeout=None):
+                 auth_plugin_map={}, read_timeout=None, write_timeout=None,
+                 bind=None):
```
methane: I prefer bind_address rather than bind. (Same to option name, translated '-' with '_')

pymysql/connections.py
```diff
@@ -640,6 +645,9 @@ def _config(key, arg):
         self.password = password or ""
         self.db = database
         self.unix_socket = unix_socket
+        if bind is not None and _py_version == (2, 6):
+            raise NotImplementedError("bind before connect is not supported in Python 2.6")
```
methane: Python 2.6 is not supported anymore. So this check is not required.

methane:
thanks

PR #545, opened 2017-02-02
Allow semicolon at the end of bulk insert query

ligyxy:
When running bulk insert with executemany(), the regex should allow semicolon at the end of the query.

### DIFF
--- a/pymysql/cursors.py
+++ b/pymysql/cursors.py
@@ -14,7 +14,7 @@ from . import err
 RE_INSERT_VALUES = re.compile(
     r"\s*((?:INSERT|REPLACE)\s.+\sVALUES?\s+)" +
     r"(\(\s*(?:%s|%\(.+\)s)\s*(?:,\s*(?:%s|%\(.+\)s)\s*)*\))" +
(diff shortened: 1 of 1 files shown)


pymysql/cursors.py
```diff
@@ -14,7 +14,7 @@
 RE_INSERT_VALUES = re.compile(
     r"\s*((?:INSERT|REPLACE)\s.+\sVALUES?\s+)" +
     r"(\(\s*(?:%s|%\(.+\)s)\s*(?:,\s*(?:%s|%\(.+\)s)\s*)*\))" +
-    r"(\s*(?:ON DUPLICATE.*)?)\Z",
+    r"(\s*(?:ON DUPLICATE.*)?)(;*\s*)\Z",
```
methane: Why do you use group?

methane:
Generally speaking, I don't want to support full SQL.  I hate complex regular expression too.
If someone want to use executemany()'s bulk insert, they should fix SQL to match the pattern.

But in this case, allowing semicolon seems simple enough to accept.

ligyxy:
You are right, no need for group.

PR #552, opened 2017-02-27
Enable reading SSL parameters from configuration file

JAORMX:
This enables the reading of some basic SSL parameters from the
configuration file. It gives precedence to the ssl parameters provided
by the class parameter (the ssl dict).

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -604,14 +604,6 @@ class Connection(object):
         if self._local_infile:
             client_flag |= CLIENT.LOCAL_FILES
 
-        self.ssl = False
-        if ssl:
-            if not SSL_ENABLED:
-                raise NotImplementedError("ssl module not found")
-            self.ssl = True
-            client_flag |= CLIENT.SSL
-            self.ctx = self._create_ssl_ctx(ssl)
-
         if read_default_group and not read_default_file:
             if sys.platform.startswith("win"):
                 read_default_file = "c:\\my.ini"
@@ -641,6 +633,21 @@ class Connection(object):
             port = int(_config("port", port))
             bind_address = _config("bind-address", bind_address)
             charset = _config("default-character-set", charset)
+            if not ssl:
+                ssl = {}
+            if isinstance(ssl, dict):
+                for key in ["ca", "capath", "cert", "key", "cipher"]:
+                    value = _config("ssl-" + key, ssl.get(key))
+                    if value:
+                        ssl[key] = value
+
+        self.ssl = False
+        if ssl:
+            if not SSL_ENABLED:
+                raise NotImplementedError("ssl module not found")
+            self.ssl = True
+            client_flag |= CLIENT.SSL
+            self.ctx = self._create_ssl_ctx(ssl)
 
         self.host = host or "localhost"
         self.port = port or 3306


pymysql/connections.py
```diff
@@ -641,6 +633,31 @@ def _config(key, arg):
...
+                "cipher": _config("ssl-cipher", ssl.get("cipher")),
+            }
+            # Update parameters, but parameters coming from the class parameter
+            # must have preference.
+            for key, value in ssl_config_params.items():
+                ssl.setdefault(key, value)
+            # No SSL parameters were given either via the class parameter or
+            # the configuration file
+            if not any(x for x in ssl.values()):
+                ssl = None
```
methane: `if not any(ssl.values())`

And `read_default_file` option changes how `ssl = {"ca": ""}` is considered.
if `read_default_file=False`, ssl is overwritten to None.  Otherwise, passed `ssl` is used as-is.
I don't like it.

pymysql/connections.py
```diff
@@ -641,6 +633,31 @@ def _config(key, arg):
...
+
+        self.ssl = False
+        if ssl:
```
methane: Mabe, `if ssl and any(ssl.values())`.

pymysql/connections.py
JAORMX: Actually, this is for the if read_default_file: case. So it will only be read when that's set.

JAORMX:
@methane I updated the commit. What do you think?

pymysql/connections.py
```diff
@@ -641,6 +633,28 @@ def _config(key, arg):
...
+                "capath": _config("ssl-capath", ssl.get("capath")),
+                "cert": _config("ssl-cert", ssl.get("cert")),
+                "key": _config("ssl-key", ssl.get("key")),
+                "cipher": _config("ssl-cipher", ssl.get("cipher")),
+            }
+            # Update parameters, but parameters coming from the class parameter
+            # must have preference.
+            for key, value in ssl_config_params.items():
+                if value:
+                    ssl.setdefault(key, value)
```
methane: since you used `_config` when building ssl_config_params,
you can just `ssl[key] = value`

JAORMX: Using    ssl[key] = value   would overwrite the value that's already there, and this gives priority to the values that are set via the ssl dict.

JAORMX: That's why I added the comment above this for loop.

methane: read `_config()` function.

JAORMX:
@methane you are right, I changed it accordingly.

pymysql/connections.py
```diff
@@ -641,6 +633,26 @@ def _config(key, arg):
...
+            ssl_config_params = {
+                "ca": _config("ssl-ca", ssl.get("ca")),
+                "capath": _config("ssl-capath", ssl.get("capath")),
+                "cert": _config("ssl-cert", ssl.get("cert")),
+                "key": _config("ssl-key", ssl.get("key")),
+                "cipher": _config("ssl-cipher", ssl.get("cipher")),
+            }
+            for key, value in ssl_config_params.items():
+                if value:
+                    ssl[key] = value
```
methane: How about this?

```
for key in ["ca", "capath", "cert", "key", "cipher"]:
    value = cfg.get(read_default_group, "ssl-" + key)
    if value:
        ssl[key] = value
```

(I hadn't check mysql options yet.)

methane: I confirmed [options](https://dev.mysql.com/doc/refman/5.7/en/mysql-command-options.html) have exactly same name.

JAORMX:
@methane by the way, I see that in the _create_ssl_ctx function there is a check such as this:
        if isinstance(sslp, ssl.SSLContext):
            return sslp
should I take that into account as well?

methane:
Ah, yes please.  Use config only when dict is None or `isinstance(ssl, dict)`

JAORMX:
@methane done; and I really liked your suggestion. I added the minimized loop you suggested as part of the change.

JAORMX:
Tested it successfully in my deployment. @methane thanks for the great reviews. Let me know if there's anything else I should change.

PR #553, opened 2017-02-28
Prevent a close() on a NoneType

blikjeham:
If _write_bytes fails, _force_close is called which sets self._sock = None. Since we can't call close() on a NoneType, we first check if the attribute is present at all.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -735,11 +735,7 @@ class Connection(object):
         except Exception:
             pass
         finally:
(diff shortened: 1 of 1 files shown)


dmulter:
Any expectation on when a release will go out with this merged change? I get this error probably once a day.

methane:
@dmulter https://groups.google.com/d/msg/pymysql-users/Lz0iuT2oJq0/8DJO43QWBAAJ

PR #555, opened 2017-03-10
ability to check cert against hostnames other than connect host

jkarneges:
We use HAProxy between PyMySQL and our MySQL server to assist with failover. This works fine for non-SSL connections, but becomes a problem when SSL is used because the hostname of the proxy ("localhost", in our case) doesn't match the hostname in the certificate presented by the MySQL server.

This type of SSL issue is usually worked around in other client software by making it possible for users to specify the connect host and logical host separately. So that's what this PR does. If `cert_hosts` is specified in the `ssl` dictionary, then the `host` will be treated as the connect host only, and the certificate will be matched against the list of values in `cert_hosts`.

Being able to check against *multiple* hosts is probably not so common (in fact I can't even think of seeing it elsewhere), but this makes it possible for HAProxy to balance connections to MySQL read replicas that may each have different certs (such as with Amazon RDS) and it's not possible for the user to know in advance which replica will be used.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -647,6 +647,7 @@ class Connection(object):
                 raise NotImplementedError("ssl module not found")
             self.ssl = True
             client_flag |= CLIENT.SSL
+            self.cert_hosts = ssl.get('cert_hosts')
             self.ctx = self._create_ssl_ctx(ssl)
 
         self.host = host or "localhost"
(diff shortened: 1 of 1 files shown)


methane:
Is this same behavior to mysqlclient?

jkarneges:
As far as I can tell, the official mysql client is unable to do this. Its default behavior is to not check the certificate host at all, and optionally it can enforce whether the connect host matches the cert.

methane:
Sorry, I can't maintenance feature to support such special setup.

jkarneges:
What would you think about letting the user specify a callback function to be run right after the connection is established? Then such custom logic could live outside of PyMySQL.

methane:
I dislike options.  Connection has too many options already.
Many people gave up about reading docstring and ask a question on issue tracker.

One possibility is splitting method and make it easy to override.

methane:
You can pass hostname and setup local DNS or write it to `/etc/hosts` to resolve the hostname to IP.

Issue #520, opened 2016-10-05
dates are returned as None


albertz:
I have a table with a column of type date with entries like `2016-00-00` (only partially set, e.g. the year only). pymysql fails to return this.

```
import pymysql
conn = pymysql.connect(...)
cur = conn.cursor()
cur.execute("SELECT publish_date FROM pub")
for row in cur:
    print(row)
```

Will print only `(None,)`.

I also [reported this for peewee](https://github.com/coleifer/peewee/issues/1086) although it looks like this is a bug on pymysql side but I'm not exactly sure.

A work around is to use

```
cur.execute("SELECT CAST(publish_date AS char) FROM pub")
```

which will return the date as a string (e.g. it prints `("2016-00-00",)`).

methane:
I haven't know MySQL allows such a broken date...
But Python's date doesn't allow such date.
What PyMySQL can do for everyone, not only for you?

```
Python 3.5.2 (default, Sep 28 2016, 18:08:09)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.38)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import datetime
>>> datetime.date(2016, 0, 0)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: month must be in 1..12
```

albertz:
Maybe, as a simple solution for now, just return it as a string in that case?

Maybe there are also other datetime-like libraries which support a data structure which allows this, but I don't know any and this would add another dependency so that might not be an option.

methane:
Hm, But doesn't it break compatibility?

Some existing code (ORMs) may not expect string is returned for date column.
But all existing code should support it, since there are NULLABLE column...

Additionally, is the date is supported by MySQL when sql_mode is strict?

methane:
```
mysql> show variables like 'sql_mode';
+---------------+-------------------------------------------------------------------------------------------------------------------------------------------+
| Variable_name | Value                                                                                                                                     |
+---------------+-------------------------------------------------------------------------------------------------------------------------------------------+
| sql_mode      | ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION |
[...]

PR #556, opened 2017-03-13
Add documentation about type conversion.

philgyford:
Including how to disable conversion for specific types.

I *think* what I've said is correct but do make sure that I haven't got something wrong!

I added a whole new page for this, but happy to organise it differently if you have a better idea in mind.

For https://github.com/PyMySQL/PyMySQL/issues/520

### DIFF
--- a/docs/source/user/conversion.rst
+++ b/docs/source/user/conversion.rst
@@ -1,0 +1,97 @@
+.. _conversion:
+
+===============
+Data Conversion
+===============
+
+By default PyMySQL will convert MySQL data types into native Python types. The
+full list of conversions is:
+
+=============== =============================================
+MySQL type      Python type
+=============== =============================================
+``BIT``         [No conversion]
+``TINY``        ``Int``
+``SHORT``       ``Int``
+``LONG``        ``Int``
+``FLOAT``       ``Float``
+``DOUBLE``      ``Float``
+``LONGLONG``    ``Int``
+``INT24``       ``Int``
+``YEAR``        ``Int``
+``TIMESTAMP``   ``datetime.datetime`` or ``datetime.date`` *
+``DATETIME``    ``datetime.datetime`` *
+``TIME``        ``datetime.timedelta`` *
+``DATE``        ``datetime.date`` *
+``SET``         ``Set``
+``BLOB``        ``String``
+``TINY_BLOB``   ``String``
+``MEDIUM_BLOB`` ``String``
+``LONG_BLOB``   ``String``
+``STRING``      ``String``
+``VAR_STRING``  ``String``
+``VARCHAR``     ``String``
+``DECIMAL``     ``Decimal``
+``NEWDECIMAL``  ``Decimal``
+=============== =============================================
+
+Items marked * are returned as ``None`` if they are of an invalid format or are
+invalid dates. For example, these valid ``DATETIME`` values::
+
+    '2007-02-25 23:06:20'
+    '2007-02-25T23:06:20'
+
+would both be converted to the following Python object::
+
+    datetime.datetime(2007, 2, 25, 23, 6, 20)
+
+But these ``DATETIME`` values would both be converted to ``None``::
+
+    '2007-02-31T23:06:20'  # There is no 31st February.
+    '0000-00-00 00:00:00'
+
+This can mean that values accepted as valid by MySQL are converted to
+``None``. For example ``DATE`` values such as these can be valid in MySQL, but
+will be converted to ``None`` by PyMySQL::
+
+   '2017-02-00'
+   '2017-00-00'
+   '0000-00-00'
+
+See `this StackOverflow answer <http://stackoverflow.com/a/6882884/250962>`_ for
+more detail on MySQL's handling of dates/times like these.
+
+If you know your database contains such dates you may want to prevent PyMSQL
+converting them (see below).
+
+Preventing conversion
+---------------------
+
+PyMySQL's automatic type conversion can be altered when establishing the
+initial connection to the database.
+
+For example, to prevent the conversion of date and time fields into native
+Python ``datetime`` types, you would delete the appropriate MySQL types from
+the dict of converters like this::
+
+    from pymysql.constants import FIELD_TYPE
+    from pymysql.converters import conversions as conv
+
+    conv = conv.copy()
+
+    del conv[FIELD_TYPE.DATE]
+    del conv[FIELD_TYPE.DATETIME]
+    del conv[FIELD_TYPE.TIME]
+
+    conn = pymysql.connect(..., conv=conv)
+
+i.e., connect as you would normally but pass in a modified dict of field types
+and conversions.
+
+The field types deleted will be returned as Strings, rather than being
+converted into any other native Python types.
+
+See the table above for the possible MySQL field types, or the ``decoders``
+dict in `pymysql/converters.py <https://github.com/PyMySQL/PyMySQL/blob/master/pymysql/converters.py>`_.
+
+
--- a/docs/source/user/index.rst
+++ b/docs/source/user/index.rst
@@ -10,5 +10,6 @@ the library as a developer.
 
   installation
   examples
+  conversion
   resources
   development


docs/source/user/conversion.rst
```diff
@@ -0,0 +1,97 @@
...
+Data Conversion
+===============
+
+By default PyMySQL will convert MySQL data types into native Python types. The
+full list of conversions is:
+
+=============== =============================================
+MySQL type      Python type
+=============== =============================================
+``BIT``         [No conversion]
```
methane: What `[No conversion]` means?  Maybe, bytes are better.

docs/source/user/conversion.rst
```diff
@@ -0,0 +1,97 @@
...
+``TINY``        ``Int``
```
methane: Please use ``int`` instead of ``Int`` for Python type.

docs/source/user/conversion.rst
```diff
@@ -0,0 +1,97 @@
...
+
+===============
+Data Conversion
+===============
+
+By default PyMySQL will convert MySQL data types into native Python types. The
+full list of conversions is:
+
+=============== =============================================
+MySQL type      Python type
```
methane: What "MySQL type" means?
It's not directly relating to [MySQL Data types](https://dev.mysql.com/doc/refman/5.7/en/data-types.html).
It's actually [Query Response's column types](https://dev.mysql.com/doc/internals/en/com-query-response.html#column-type).

docs/source/user/conversion.rst
```diff
@@ -0,0 +1,97 @@
...
+``TIME``        ``datetime.timedelta`` *
+``DATE``        ``datetime.date`` *
+``SET``         ``Set``
+``BLOB``        ``String``
+``TINY_BLOB``   ``String``
+``MEDIUM_BLOB`` ``String``
+``LONG_BLOB``   ``String``
+``STRING``      ``String``
+``VAR_STRING``  ``String``
+``VARCHAR``     ``String``
```
methane: All of them are str or bytes.

docs/source/user/conversion.rst
```diff
@@ -0,0 +1,97 @@
...
+``DECIMAL``     ``Decimal``
+``NEWDECIMAL``  ``Decimal``
+=============== =============================================
+
+Items marked * are returned as ``None`` if they are of an invalid format or are
+invalid dates. For example, these valid ``DATETIME`` values::
```
methane: While it's current behavior, I think it is implementation detail, and I'm thinking about changing this
to return string instead of None.
Documenting it makes harder to change it.  Maybe, I won't change if it's documented behavior.

docs/source/user/conversion.rst
```diff
@@ -0,0 +1,97 @@
...
+will be converted to ``None`` by PyMySQL::
+
+   '2017-02-00'
+   '2017-00-00'
+   '0000-00-00'
+
+See `this StackOverflow answer <http://stackoverflow.com/a/6882884/250962>`_ for
+more detail on MySQL's handling of dates/times like these.
+
+If you know your database contains such dates you may want to prevent PyMSQL
```
methane: s/PyMSQL/PyMySQL/

docs/source/user/conversion.rst
```diff
@@ -0,0 +1,97 @@
+.. _conversion:
+
+===============
+Data Conversion
+===============
+
+By default PyMySQL will convert MySQL data types into native Python types. The
+full list of conversions is:
```
philgyford: Good point; I'll add a separate table based on the list of `encoders`.

docs/source/user/conversion.rst
philgyford: So, just "bytes"? Or `bytes`?

docs/source/user/conversion.rst
philgyford: So, like:

> `str` or bytes

?

docs/source/user/conversion.rst
philgyford: I think it's useful to mention the issue with invalid dates not being converted to native python date/time types, as it took me a while to figure out what was happening when this happened to me.

Instead of removing this section we could say that this behaviour is likely to change in the future, to return strings (which sounds a good idea) rather than None?

methane: Please add note about (sorry, my English is bad,  please rewrite to more good sentence):

* Customizing converter is **hack** and **not recommended**.  It's integrated implementation detail very tightly, so it's difficult to keep backward compatibility sometimes.
*  You should try converting at higher layer like ORM always when possible.
*  If your DB contains broken data, you should fix it, while you can use converter hack for temporal workaround.

docs/source/user/conversion.rst
```diff
@@ -0,0 +1,97 @@
+.. _conversion:
+
+===============
+Data Conversion
+===============
+
```
methane: Please add these notes at top:

* Information in this page is implementation detail.  Normal PyMySQL users don't need this information.
* This is backdoor for low level problem.  Don't use this for (de)serializing application layer types.  Use ORM instead.

docs/source/user/conversion.rst
methane: ` ``bytes`` `

docs/source/user/conversion.rst
methane: yes.

docs/source/user/conversion.rst
philgyford: If you'd prefer not to have this page in the documentation at all that's fine :) I thought it might be useful, but if you think it's going to cause you difficulties then we can just leave it alone? Let me know.

methane: I don't will wasting your work.  I just want strong message "don't use this, unless no other way!" in this document.

This feature is kept with many pitfalls and special dirty hacks.  This feature makes difficult to finding and fixing bugs without regression or backward incompatibility.

For example, see PyMySQL/mysqlclient-python#145 and PyMySQL/mysqlclient-python#155 are recent regression caused by basic design of converter.

Another example is this dirty code in https://github.com/PyMySQL/mysqlclient-python/blob/master/MySQLdb/connections.py
:

```
        db = proxy(self)
        def _get_string_literal():
            # Note: string_literal() is called for bytes object on Python 3 (via bytes_literal)
            def string_literal(obj, dummy=None):
                return db.string_literal(obj)
            return string_literal
```

converter (both of encoder and decoder) should know encoding of connection.
But converters doesn't have encoding argument.  So string encoder is overwritten by connection constructor.  And weakref is required to avoid circular reference (it cause leak at Python 2.  I now hate Python 2 too...).
Decoder has another dirty hack too.

If no one use converters, I can add new argument and pass necessary information via argument.  Then I'll be able to remove lots of unmaintainable dirty hacks!!!

I want to do it someday.  But it breaks backward compatibility.
That's why I don't want to people using it.

philgyford: No problem - I'll revise things next week; no more time until then I'm afraid.

rholloway:
I'm not sure this will be as uncommon as you may think. While at first the change in #520 to convert to strings sounds ok, it actually caused issues for us.  We also had many rows with 0000-00-00 00:00:00  in a datetime field in DB which is fine in MySQL but not in python.

Up through 0.7.11, this was not an issue as they were handled as None through the stack. This is similar to how other mysql drivers function, fwiw (returning None in such cases).

I tested an upgrade to 0.8.0 and the change modifies data coming back not only by not returning None in cases where it used to, but returning a different type altogether than what you would expect (a string rather than datetime).

We can certainly account for it application side and add other checks, but defaulting to this change in 0.8.0 without appropriate documentation does indeed cause issues, especially given ORMs such as SQLAlchemy don't account for it and handle either (as of the latest version of SQLAlchemy, you can define your attribute as datetime but it'll still come back a string if using pymysql 0.8.0)

methane:
> I'm not sure this will be as uncommon as you may think.

What the "this" means?
If you meant `conv` argument, you many underestimate how it's evil.
It waste my time a lot, may be more than 100 hours.
It made me tired about maintaining PyMySQL and mysqlclient.
It is broken by design before I start maintaining these projects.

methane:
> I meant the change in #520 that would require overriding conv or custom conversion hack as of 0.8.0.

You can update your DB to drop all invalid dates too.  Or you can convert string to None in application layer too.
So hack is not strong requirement for driver layer.

> to maintain consistency with other drivers in the ORM layer...

What *other drivers* mean?  Do you have list of behavior of drivers?

I want to make mysqlclient and PyMySQL as close as possible.
So I will change mysqlclient too.  Or I will revert #520 in PyMySQL.
You are the only person who report suffering from #520.  So I'm not sure about I should revert or not.

methane:
Hm, MySQL Connector/Python seems return None for invalid date.
https://github.com/mysql/mysql-connector-python/blob/a602f150e40970e5c72acf6a537dc6935384f930/lib/mysql/connector/conversion.py#L505-L524

jongoodrumacv:
Hi Everyone,

What is the status of this pull request?  These conversions are very important piece of documentation and I'm wondering how accurate these changes are.  Can I assume these type conversions are correct?

Thanks.

PR #558, opened 2017-03-15
easyMVC

thautwarm:
I think it a wrapper of PyMySQL which is not only easy but also quite graceful!
For example, you can write the entities of a web project with just using jsons like below:
`
[
    {
        "class":"user",
        "attrs":["id","username","password","access","info"]
    },
    {
        "class":"course",
        "attrs":["id","name","time","finalexamdate"]
    }
]
` 
And you can write the graceful codes like these below:
`
user1=copy.deepcopy(entities.user)
user1.username='archer'
user1.password=')(*&^'

userdao.add(username='saber','password'='!@#$%^&')
userdao.add(user=user1)
`

### DIFF
--- a/pymysql/easyMVC/__init__.py
+++ b/pymysql/easyMVC/__init__.py
@@ -1,0 +1,1 @@
+from . import dao,entity,service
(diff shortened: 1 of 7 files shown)


methane:
Please release it by yourself.
I don't have enough time and energy to continue all packages I maintain.
I won't accept no additional maintenance cost.

thautwarm:
Ok.
Quite sorry if I bothered you.

Issue #362, opened 2015-07-31
Feature Request: MySQL login path support


seagoj:
I was hoping you'd consider adding login via --login-path. This would allow for PyMySQL to make connections via encrypted connection information in the ~/.mylogin.cnf file.

Because of the encryption, I think you would have to pass the parameter all the way down to the mysql client, which may not even be possible, but it would be a nice feature.

[Documentation](http://dev.mysql.com/doc/mysql-utilities/1.5/en/mysql-utils-intro-connspec-mylogin.cnf.html)

Thanks

methane:
I can't find format spec of mylogin.cnf.
I read the mysql client and found it uses AES128 for encryption.
It means it requires additional dependency for encryption libraries.
I'll consider about this after 0.7.

tsroten:
@methane Here is the implementation that mycli uses to read the login path file.

ngaitanis:
Has there been any action on this? Perhaps an estimate on when/if this will be implemented?

methane:
No action, and no estimate.

simnim:
This feature is super useful and would be great for security.
I hope it gets more attention.

methane:
> This feature is super useful and would be great for security.

NO!  The format put key and lock in one place.  All attacker can decrypt it.
It's not safer than plaintext from attacker's point of view.
It can even dangerous if people think it improve security.

I don't think it's useful.  But I know there are people feel it's useful.
I'll accept the feature if it:

* is consistent with MySQL Connector/C
* doesn't slowdown making new connection
* well tested, and easy to maintain (I never use it, contributors should maintain it)

simnim:
> NO! The format put key and lock in one place. All attacker can decrypt it.

Ok, maybe great was the wrong word, but I think "NO!" is too strong a reaction.

The encryption is shit, it's true, but that's mostly irrelevant.

I'm not trying to start a flamewar or anything, I just think there are valid reasons why it can be good for security. Mainly that you don't store passwords in your repo and you can put them in locked down locations etc.

Besides, how is it any better than `~/.ssh/id_rsa`? (I know you can add a password to it, but usually people don't for programmatic scp and such.)

Checking code into a repo that looks like this is arguably much worse.
```
# Connect to the database
connection = pymysql.connect(host='localhost',
                             user='user',
                             password='passwd',
                             db='db',
                             charset='utf8mb4',
                             cursorclass=pymysql.cursors.DictCursor)
```

[...]

PR #561, opened 2017-03-22
Add encrypted login path file support.

tsroten:
This addresses #362.

This pull request adds an optional dependency, the [cryptography package](https://github.com/pyca/cryptography), which supports Python 2.6-2.7, Python 3.3+, and PyPy 2.6+. This package is used to decrypt the encrypted login path file so that `host`, `user`, `password`, `socket`, and `port` can be read from a login path group.

If users want to install PyMySQL with login path support, they simply run `pip install pymysql[loginpath]`, which will install both PyMySQL and cryptography.

The `Connection()` keyword arguments for `host`, `user`, `password`, `unix_socket`, and `port` continue to have the highest precedence, so they outweigh values in the login path file.

The login path file is expected to be at:
- `%APPDATA%\MySQL\.mylogin.cnf` (Windows)
- `~/.mylogin.cnf` (non-Windows)

Or, the `MYSQL_TEST_LOGIN_FILE` environment variable is read for an alternative login path file location. This is per the [MySQL docs](https://dev.mysql.com/doc/refman/5.7/en/option-files.html).

If `Connection()` is called with the `login_path` argument, but cryptography is not installed, then an `ImportError` is raised with an appropriate message.

If the `login_path` argument is passed to `Connection()`, but the login path file is unable to be opened, then a `RuntimeError` is raised.

## Installation
Either of these will install PyMySQL with login path file support:
- `pip install pymysql[loginpath]`
- `pip install pymysql cryptography`

## Examples

```python
from pymysql.connections import Connection

# Login path file is not opened and cryptography is not imported.
Connection(user=user, host=host, password=password)

# Login path file is opened and `client` group is read.
Connection(login_path=True)

# Login path file is opened and `qa1` group is read.
Connection(login_path='qa1')

# Login path file is opened, `client` group is read, but `host` is overriden.
Connection(host='db.domain.com', login_path='client')
```

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -536,7 +536,7 @@ class Connection(object):
                  autocommit=False, db=None, passwd=None, local_infile=False,
                  max_allowed_packet=16*1024*1024, defer_connect=False,
                  auth_plugin_map={}, read_timeout=None, write_timeout=None,
-                 bind_address=None):
+                 bind_address=None, login_path=None):
         """
         Establish a connection to the MySQL database. Accepts several
         arguments:
@@ -583,6 +583,8 @@ class Connection(object):
             The class needs an authenticate method taking an authentication packet as
             an argument.  For the dialog plugin, a prompt(echo, prompt) method can be used
             (if no authenticate method) for returning a string from the user. (experimental)
+        login_path: A login path group to read from the encrypted login path
+            file. Or, `True` to read the default "client" group.
         db: Alias for database. (for compatibility to MySQLdb)
         passwd: Alias for password. (for compatibility to MySQLdb)
         """
@@ -604,6 +606,28 @@ class Connection(object):
         if self._local_infile:
             client_flag |= CLIENT.LOCAL_FILES
 
+        if login_path:
+            if login_path is True:
+                login_path = 'client'
+
+            from .login_path import open_login_path_file
+            login_path_file = open_login_path_file()
+
+            if (login_path_file is None):
+                raise RuntimeError('Error reading login path file.')
+
+            lp_cfg = Parser()
+            if PY2:
+                lp_cfg.readfp(login_path_file)
+            else:
+                lp_cfg.read_file(login_path_file)
+
+            user = lp_cfg.get(login_path, 'user', user)
+            password = lp_cfg.get(login_path, 'password', password)
+            host = lp_cfg.get(login_path, 'host', host)
+            unix_socket = lp_cfg.get(login_path, 'socket', unix_socket)
+            port = int(lp_cfg.get(login_path, 'port', port))
+
         if read_default_group and not read_default_file:
             if sys.platform.startswith("win"):
                 read_default_file = "c:\\my.ini"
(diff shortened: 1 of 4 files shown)


pymysql/connections.py
```diff
@@ -583,6 +583,8 @@ def __init__(self, host=None, user=None, password="",
             The class needs an authenticate method taking an authentication packet as
             an argument.  For the dialog plugin, a prompt(echo, prompt) method can be used
             (if no authenticate method) for returning a string from the user. (experimental)
+        login_path: A login path group to read from the encrypted login path
+            file. Or, `True` to read the default "client" group.
```
methane: > Or, `True` to read the default "client" group.

I don't like it.
Is this consistent with MySQL Connector/C?

pymysql/login_path.py
```diff
@@ -0,0 +1,122 @@
...
+    try:
+        with open(path, 'rb') as fp:
+            key = read_key(fp)
+            cipher = get_aes_cipher(key)
+            plaintext = decrypt_file(fp, cipher.decryptor())
+    except (OSError, IOError):
+        return None
+
+    if not isinstance(plaintext, BytesIO):
+        return None
```
methane: How is this happen?

pymysql/login_path.py
```diff
@@ -0,0 +1,122 @@
...
+from io import BytesIO, TextIOWrapper
+import os
+import struct
+
+try:
+    from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+    from cryptography.hazmat.backends import default_backend
+except ImportError:
+    raise ImportError('You must install the package "cryptography" in order '
+                      'to read the login path file.')
```
methane: I don't like this.  Use flag.

```py
try:
    from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
    from cryptography.hazmat.backends import default_backend
    HAVE_CRYPTOGRAPHY = True
except ImportError:
    HAVE_CRYPTOGRAPHY = False
```

pymysql/connections.py
```diff
@@ -604,6 +606,28 @@ def __init__(self, host=None, user=None, password="",
         if self._local_infile:
             client_flag |= CLIENT.LOCAL_FILES
 
+        if login_path:
+            if login_path is True:
+                login_path = 'client'
+
+            from .login_path import open_login_path_file
+            login_path_file = open_login_path_file()
```
methane: I don't like this design.
Please move parsing code to login_path module too.

```
opts = login_path.read()
# opts is dict if successfully read option file.
# opts is None if cryptography is not available, or failed to open/read/decrypt mylogin.cnf
```

With this API, people can test login_path file easily, without creating real connection.

pymysql/login_path.py
```diff
@@ -0,0 +1,122 @@
...
+
+# Number of bytes used to store the length of ciphertext.
+CIPHER_STORE_LENGTH = 4
+
+
+def open_login_path_file():
+    """Open a decrypted version of the login path file."""
+    path = get_login_path_file()
+    try:
+        with open(path, 'rb') as fp:
```
methane: See https://github.com/mysql/mysql-server/blob/23032807537d8dd8ee4ec1c4d40f0633cd4e12f9/mysys_ssl/my_default.cc#L1496-L1520

Without it, this is insecure.

pymysql/login_path.py
```diff
@@ -0,0 +1,122 @@
...
+
+def create_key(key):
+    """Create the AES key from the login path file header."""
+    rkey = [0] * 16
+    for i in range(len(key)):
+        try:
+            rkey[i % len(rkey)] ^= ord(key[i:i + 1])
+        except TypeError:
+            # ord() was unable to get the value of the byte.
+            return None
```
methane: When is this happen?
If it shouldn't happen, don't catch exception.

tsroten:
@methane Thanks for the feedback. I'll try to update the pull request soon.

methane:
I decided to not implement this to avoid increasing complexity and maintenance cost.
I lost motivation to grow this project already.

Instead, I've released [myloginpath](https://pypi.org/project/myloginpath/), based on your pull request.
Thank you for your contribution.

PR #570, opened 2017-05-15
latin1 as default seriosly?

SKYnv:
It's solve many problems with non latin charsets


methane:
You can pass charset parameter.
latin1 is the default of MySQL Connector/C.

When MySQL Connector/C changes the default, PyMySQL follows. Not now.

Issue #571, opened 2017-05-18
Out of date API docs


smitchell556:
The API docs seem to be out of date for `connections.py` and `cursors.py` and don't include connection and cursor properties or function definitions.

It looks like the documentation started from #199 but didn't quite get finished.

methane:
I won't work on it.  But contribution is welcome.

smitchell556:
Do you prefer just class level documentation as it currently is? I think adding class members would be beneficial.

methane:
Do you know [PEP 249](https://www.python.org/dev/peps/pep-0249/)?

Keyword arguments of connection constructor is PyMySQL specific.  So they are documented.
On the other hand, most methods are based on PEP 249.
And most methods not specified in PEP 249 is implementation detail.

smitchell556:
I am familiar with it. I know the PEP is referenced in the user guide, but adding it to the main API reference page is a good idea.

The arguments accepted by the connection constructor could be moved into the class docstring so they would show up in the pages generated on readthedocs. As it stands sphinx doesn't use the __init__ docstring if there is a class level docstring. There are also some public members that aren't PEP 249 specific, like connection.open which I think would be beneficial to document. I'll submit a PR and we can go from there.

PR #573, opened 2017-05-21
Update API docs

smitchell556:
I added `DictCursor` and `SSDictCursor` to docs and moved the `__init__` docstring for the `Connection` class into the class level docstring, otherwise sphinx won't include the parameter descriptions in the build.

I also added the root path of the directory to the system path in `conf.py`. I was unable to build the docs locally without it and it's necessary for `autodoc` to work.

Related to #571

### DIFF
--- a/docs/source/modules/connections.rst
+++ b/docs/source/modules/connections.rst
@@ -7,5 +7,5 @@ Connection Object
    :members:
    :exclude-members: DataError, DatabaseError, Error, InterfaceError,
                      IntegrityError, InternalError, NotSupportedError,
-		     OperationalError, ProgrammingError, Warning,
-		     escape, literal, write_packet
+                     OperationalError, ProgrammingError, Warning,
+                     escape, literal, write_packet
--- a/docs/source/modules/connections.rst
+++ b/docs/source/modules/connections.rst
@@ -7,4 +7,5 @@ Connection Object
    :members:
    :exclude-members: DataError, DatabaseError, Error, InterfaceError,
                      IntegrityError, InternalError, NotSupportedError,
-                     OperationalError, ProgrammingError, Warning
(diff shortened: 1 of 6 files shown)


docs/source/modules/connections.rst
```diff
@@ -1,6 +1,10 @@
...
+=================
 
-.. automodule:: pymysql.connections
+.. module:: pymysql.connections
 
 .. autoclass:: Connection
+   :members:
+   :exclude-members: DataError, DatabaseError, Error, InterfaceError,
+                     IntegrityError, InternalError, NotSupportedError,
+                     OperationalError, ProgrammingError, Warning
```
methane: exclude `escape`, `literal` and `write_packet` too.

smitchell556: No problem, it's been updated.

docs/source/modules/connections.rst
```diff
@@ -7,4 +7,5 @@ Connection Object
    :members:
    :exclude-members: DataError, DatabaseError, Error, InterfaceError,
                      IntegrityError, InternalError, NotSupportedError,
-                     OperationalError, ProgrammingError, Warning
+		     OperationalError, ProgrammingError, Warning,
+		     escape, literal, write_packet
```
methane: did you use TAB for indent?

smitchell556: Hmm that's weird. I use autopep8 to avoid these problems, but maybe I had turned it off for some reason. Will fix when I get to the office sorry.

smitchell556: Now that I think about it, it's a`.rst` file so autopep8 wouldn't do anything. That's probably why it ended up that way because autopep8 didn't kick in like I expected.

methane:
thanks

PR #575, opened 2017-05-26
Fix `TypeError` with Python 2.7

rjollos:
TypeError: 'encoding' is an invalid keyword argument for this function


rjollos:
Here's the error I was seeing:

```sh
$ python --version
Python 2.7.13
$ python setup.py develop
  File "setup.py", line 12, in <module>
    with open('./README.rst', encoding='utf-8') as f:
TypeError: 'encoding' is an invalid keyword argument for this function
```

PR #576, opened 2017-05-26
Don't apppend `.None` to version number

rjollos:


rjollos:
Fix this issue:
```pycon
>>> import pymysql
>>> pymysql.__version__
'0.7.11.None'
```

PR #577, opened 2017-06-06
added more character sets though 303

jwoos:


rmn36:
Can someone please merge this? I am currently facing an issue that is being caused by this. It is a simple fix.

methane:
I can't confirm them.
Additionally, collation id is 1byte in wire protocol.  So PyMySQL can't support collations over 256.

You should start utf8mb4 and use `init_command="SET NAMES utf8mb4 COLLATE <collation you want to use>` instead.

methane:
MySQL 5.7.18:

```
...
| 246 | utf8mb4            | utf8mb4_unicode_520_ci   |            |
| 247 | utf8mb4            | utf8mb4_vietnamese_ci    |            |
| 248 | gb18030            | gb18030_chinese_ci       | Yes        |
| 249 | gb18030            | gb18030_bin              |            |
| 250 | gb18030            | gb18030_unicode_520_ci   |            |
+-----+--------------------+--------------------------+------------+
222 rows in set (0.00 sec)
```

jwoos:
Going to close this, I pulled the wrong docker image version (8) when I was working with this

methane:
Even if MySQL 8 is released, PyMySQL can't use collations which id > 255.

jwoos:
Those were the character sets that I got by running the commands that were specified

PR #578, opened 2017-06-12
KeyError 255

sah4ez:
Add to charsets default UTF-8 for 255 key.


methane:
```
mysql> select id, character_set_name, collation_name, is_default from information_schema.collations where collation_name = 'utf8_unicode_ci';
+-----+--------------------+-----------------+------------+
| id  | character_set_name | collation_name  | is_default |
+-----+--------------------+-----------------+------------+
| 192 | utf8               | utf8_unicode_ci |            |
+-----+--------------------+-----------------+------------+
1 row in set (0.00 sec)
```

PR #582, opened 2017-06-20
Fix connection procedure auth-switch wrong plugin auth method.

lishuode:
For AuthSwitch feature, mysql_native_password and mysql_old_password
should reply only 8-byte or 20-byte scrambled password without '\0'. Use '\0' will let MySQL server reply Error Code 1043 which means Bad HandShake.
Please see the https://web.archive.org/web/20170405005641/https://dev.mysql.com/doc/internals/en/secure-password-authentication.html . 
Also see connections.py line 11130, which also responsed with no '\0'.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -1186,7 +1186,7 @@ class Connection(object):
             data = _scramble(self.password.encode('latin1'), auth_packet.read_all())
         elif plugin_name == b"mysql_old_password":
             # https://dev.mysql.com/doc/internals/en/old-password-authentication.html
-            data = _scramble_323(self.password.encode('latin1'), auth_packet.read_all())
+            data = _scramble_323(self.password.encode('latin1'), auth_packet.read_all()) + b'\0'
         elif plugin_name == b"mysql_clear_password":
             # https://dev.mysql.com/doc/internals/en/clear-text-authentication.html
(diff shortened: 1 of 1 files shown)


methane:
Could you write test case for it?

lishuode:
Hi @methane , I can not write a unit test for this scenario. 
This is because the default auth plugin is only mysql_native_password and sha256_password, PyMySQL is not support sha256_password now. And auth switch took place when default auth plugin is not match the users. So I need start a new MySQL with default sha256, then create a native user. So the server say SHA256 then the client say SHA256, after that the server say native, then the client say native.

You may notice me that I recall the change of Old Passwrod plugin.  This is for the code https://github.com/mysql/mysql-server/blob/5.6/sql/sql_acl.cc#L12129 which means the change is useless. But for Native Password Plugin, this change is needed. For https://github.com/mysql/mysql-server/blob/5.6/sql/sql_acl.cc#L12088

Also, in fact, we already have a UT which contains a path native_password switch to old_password during  method testMySQLOldPasswordAuth. 
My scenario is that I use a MySQL Proxy which have a native_password switch again native_password. But this fix is need when PyMySQL support SHA256. This is a real bug fix not a work around.

artparks:
Can this be released? It seems to prevent this library connecting to Azure's hosted MySQL over SSL.

Issue #532, opened 2016-11-26
SHA256 user account issue


jmcgrath207:
Issue with user account that uses SHA256 Password.

I was trying to connect to mysql 5.7.16 docker container with a user account I created with a SHA256 password and I am getting this error.

```
  File "/usr/local/lib/python3.5/site-packages/pymysql/__init__.py", line 90, in Connect
    return Connection(*args, **kwargs)
  File "/usr/local/lib/python3.5/site-packages/pymysql/connections.py", line 688, in __init__
    self.connect()
  File "/usr/local/lib/python3.5/site-packages/pymysql/connections.py", line 906, in connect
    self._request_authentication()
  File "/usr/local/lib/python3.5/site-packages/pymysql/connections.py", line 1123, in _request_authentication
    auth_packet = self._process_auth(plugin_name, auth_packet)
  File "/usr/local/lib/python3.5/site-packages/pymysql/connections.py", line 1185, in _process_auth
    raise err.OperationalError(2059, "Authentication plugin '%s' not configured" % plugin_name)
pymysql.err.OperationalError: (2059, "Authentication plugin 'b'sha256_password'' not configured")

Process finished with exit code 1
```

Starting at line 1147 it appears this is where the decision is made based on the authentication types but I don't see any that mention SHA256

## Executable script to reproduce (for bugs)

```
#Open database connection
db = pymysql.connect("mysql","admin","password")
#prepare a cursor object using cursor() method
cursor = db.cursor()
```

## Your Environment

* Operating System and version:
PRETTY_NAME="Debian GNU/Linux 8 (jessie)"
NAME="Debian GNU/Linux"
VERSION_ID="8"
VERSION="8 (jessie)"

Linux 701fd9099c55 3.16.0-38-generic #52~14.04.1-Ubuntu SMP Fri May 8 09:43:57 UTC 2015 x86_64 GNU/Linux

mysql 5.7.16 docker container

* PyMySQL Version used:

PyMySQL==0.7.9

methane:
No plan to support sha256 password plugin.
Contribution is welcome.

lishuode:
#583

lishuode:
@sonance207 I'm adding code for SHA256 now. #583 SHA256 plugin support.

sonance207:
The issue is being resolved in #583. Closing case

sharma-dhrv:
[...]

PR #583, opened 2017-06-23
Add support for SHA256 auth plugin

elemount:
For issue #532 
SHA256 authentication plugin is the recommended auth way for the MySQL now, for security reason.
This covers all the SHA256 auth code pass.
1. mysqld with default_auth_plugin=sha256_password, and use SSL, server say sha256, client and server get SSL, and then client response sha256 password directly.
2. mysqld with default_auth_plugin=sha256_password, and do not use SSL, server say sha256, client response sha256, server (1) auth switch (2) tell the RSA public key directly, then client response the encrypted password.
3. mysqld with default_auth_plugin=mysql_native_password(By default), and use SSL, server say native, client and server get SSL, and then client response native, and server say AuthSwith, the client reponse the password directly.
4. mysqld with default_auth_plugin=mysql_native_password(By default), and do not use SSL, server say native, client say native, server say AuthSwith, and the client get RSA(already have or request '1' then server response public key) and response the encrypted password.

I test code in both 4 scenario. But can not add a proper unit test for it now. May need some help.


methane:
I don't want to add dependency

lishuode:
@methane I use 
```{python}
try:
    from cryptography.hazmat.backends import default_backend
    from cryptography.hazmat.primitives import serialization, hashes
    from cryptography.hazmat.primitives.asymmetric import padding
    HAVE_CRYPTOGRAPHY = True
except ImportError:
    HAVE_CRYPTOGRAPHY = False
```
to make sure this module is not required.

methane:
Can you use `hashlib`'s sha256?

elemount:
Sure, let me check this lib

_Sent from my Xiaomi MI 5 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

elemount:
Sure, let me check this lib.

elemount:
@methane , I'm afraid of that I can not use `hashlib` instead of cryptography.
The `sha256_password` auth method is not a hash algorithm to encrypt secret, it actually use RSA to encrypted password, in detail, it need to use [RSA_PKCS1_OAEP_PADDING](https://wiki.openssl.org/index.php/Manual:RSA_public_encrypt(3)) algorithm to encrypt the password.
In `hashlib` lib, its all available algorithm is 
```
In [2]: hashlib.algorithms_available
Out[2]: 
{'DSA',
 'DSA-SHA',
 'MD4',
 'MD5',
 'RIPEMD160',
 'SHA',
 'SHA1',
 'SHA224',
 'SHA256',
 'SHA384',
 'SHA512',
 'dsaEncryption',
 'dsaWithSHA',
 'ecdsa-with-SHA1',
 'md4',
 'md5',
 'ripemd160',
 'sha',
 'sha1',
 'sha224',
 'sha256',
 'sha384',
 'sha512',
 'whirlpool'}
``` 
which do not have any algorithm which is related with RSA.
If you do not like `cryptography`, I can use `PyCrypto` instead of `cryptography`, or any other lib which you propose.

methane:
> If you do not like cryptography, I can use PyCrypto instead of cryptography, or any other lib which you propose.
> View changes

I don't dislike cryptography.  I just don't want **any** dependency.

methane:
As far as I read the manual, RSA is not required when tls is used.
And binary distribution of MySQL Community (GPL) Edition doesn't support RSA.

methane:
And I don't want to add any new code to `connections.py`.  It's toooo large already.
Please move your code to `pymysql/auth/sha256password.py`.

elemount:
@methane You are right, RSA works when the connection is not using SSL. And only built with OpenSSL but not YaSSL(which is default), this mechanism will works. I'll make sure that when using SSL, the exception will not thrown, and when the MySQL server communicate with RSA, it will throw the module needed, or you have any other suggestion?

For the connection.py code is too heavy, sure, let me move it to pymysql/auth/sha256password.py.

pymysql/connections.py
```diff
@@ -144,6 +151,21 @@ def _scramble(password, message):
...
 
+def _xor_password(password, salt):
+    password_bytes = bytearray(password, 'ascii')
+    for i in range(len(password_bytes)):
+        password_bytes[i] ^= ord(salt[i % len(salt)])
+    return password_bytes.decode('ascii')
+
+def _sha256_rsa_crypt(password, salt, public_key):
+    if not HAVE_CRYPTOGRAPHY:
+        raise NotImplementedError("cryptography module not found")
```
methane: Please use OperationalError instead of NotImplementedError.

pymysql/connections.py
```diff
@@ -144,6 +151,21 @@ def _scramble(password, message):
     result = s.digest()
     return _my_crypt(result, stage1)
 
+def _xor_password(password, salt):
+    password_bytes = bytearray(password, 'ascii')
+    for i in range(len(password_bytes)):
+        password_bytes[i] ^= ord(salt[i % len(salt)])
```
methane: Write `salt_len = len(salt)` before loop and use it.

pymysql/connections.py
```diff
@@ -144,6 +151,21 @@ def _scramble(password, message):
...
+    return password_bytes.decode('ascii')
```
methane: This may cause UnicodeDecodeError.  Keep bytearray type as-is.

pymysql/connections.py
```diff
@@ -374,6 +396,9 @@ def is_auth_switch_request(self):
         # http://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::AuthSwitchRequest
         return self._data[0:1] == b'\xfe'
 
+    def is_rsa_public_key(self):
+        return self._data[0:1] == b'\x01'
```
methane: Is `\x01` really RSA_PUBLIC_KEY packet?
Isn't it a more general purpose packet which has other semantics in other context?

methane:
Since Oracle removes server-client protocol document, it's hard to review protocol...
Why Oracle remove such an very important document....

pymysql/connections.py
elemount: Sure.

pymysql/connections.py
elemount: Sure.

pymysql/connections.py
elemount: Sure.

pymysql/connections.py
elemount: I think it is more general purpose packet, but I can not find a proper name, I'll fix it by the next commit for a proper packet name.

elemount:
I use this web archived document https://web.archive.org/web/20170404163410/https://dev.mysql.com/doc/internals/en/sha256.html . 
This is another link which is helpful https://dev.mysql.com/doc/refman/5.7/en/sha256-pluggable-authentication.html .

methane:
I know webarchive, but all images (including important sequence diagram) gone.

pymysql/connections.py
elemount: I found the real meaning of `\x01`, it means  [Extra Authentication Data](https://web.archive.org/web/20170404163312/https://dev.mysql.com/doc/internals/en/successful-authentication.html).

elemount:
@methane Could you help to review it again?

elemount:
Hi @methane ,I squashed this PR and fixed up the tests. Could you help to review it?

sscherfke:
You can use [extras_require](http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) in `setup.py` to declare `cryptography` as optional dependency, e.g.,
```
setup(
    name="PyMySQL",
    # ...
    extras_require={
        'auth_sha256': ['cryptography>=2.1'],
    },
    classifiers=[
        # ...
    ],
)
```
Users can then `pip install pymysql` w/o dependenies and `pip install pymysql[auth_sha256]` with dependencies.

methane:
@elemount Why did you close this?  May I continue your work?

elemount:
I changed it because I abandoned using SHA256 in my MySQL server. 
You can continue my work with current code. I believe these code can work.

grooverdan:
@elemount, thanks for your work. I've started to look at this. Overall looks pretty good and I've only added a couple of Py3 compatibility fixes so far.
`
server_public_key` looks like raw bytes read from the server but a filename ` serialization.load_pem_public_key` when passed from to the connection. I haven't thought yet how to resolve this.

Hope the auth plugin interface wasn't too bad/buggy.

WIP: 

https://github.com/grooverdan/PyMySQL/compare/travis-docker-mysql-8.0...grooverdan:sha256_auth_plugin

Next steps:
reenable existing plugin tests (on tcp, unix_socket tests appear disabled)
write more tests (thankfully MySQL-8.0 seems to generate and configure client/server ssl keys as well).
implement the caching mechanism and check all the paths are implemented: https://dev.mysql.com/doc/dev/mysql-server/latest/page_caching_sha2_authentication_exchanges.html

Issue #584, opened 2017-06-29
Can not connect Azure Database for MySQL.


elemount:
Connect with Azure Database for MySQL got this error.

Traceback (most recent call last):
  File "example.py", line 6, in <module>
    conn = pymysql.connect(host='shuodlclient.mysql.database.azure.com', port=3306, user='mysqlaas@shuodlclient', passwd='test_PyMySQL', db='mysql')
  File "/home/lishuode/code/PyMySQL/pymysql/__init__.py", line 90, in Connect
    return Connection(*args, **kwargs)
  File "/home/lishuode/code/PyMySQL/pymysql/connections.py", line 705, in __init__
    self.connect()
  File "/home/lishuode/code/PyMySQL/pymysql/connections.py", line 931, in connect
    self._request_authentication()
  File "/home/lishuode/code/PyMySQL/pymysql/connections.py", line 1160, in _request_authentication
    auth_packet = self._process_auth(plugin_name, auth_packet)
  File "/home/lishuode/code/PyMySQL/pymysql/connections.py", line 1225, in _process_auth
    pkt = self._read_packet()
  File "/home/lishuode/code/PyMySQL/pymysql/connections.py", line 1013, in _read_packet
    packet.check_error()
  File "/home/lishuode/code/PyMySQL/pymysql/connections.py", line 393, in check_error
    err.raise_mysql_exception(self._data)
  File "/home/lishuode/code/PyMySQL/pymysql/err.py", line 107, in raise_mysql_exception
    raise errorclass(errno, errval)
pymysql.err.OperationalError: (1045, u"Access denied for user 'mysqlaas'@'167.220.255.151' (using password: YES)")

PR #585, opened 2017-06-29
Fix handle AuthSwitch packet bug.

elemount:
MySQL documents annouce the AuthSwitch packet is contains with two
component `auth_plugin_name` and `auth_data`, which `auth_data` is
a string[EOF] string. But in fact it will return a string[NUL] string
or can also say the string[EOF] is consist of a 20bytes string and a '\0'
byte. Now we just follow the document which use those 21bytes as salt
and that is not correct.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -132,6 +132,8 @@ def dump_packet(data): # pragma: no cover
     print()
 
 
+SCRAMBLE_LENGTH = 20
+
 def _scramble(password, message):
     if not password:
         return b''
@@ -139,7 +141,7 @@ def _scramble(password, message):
     stage1 = sha_new(password).digest()
     stage2 = sha_new(stage1).digest()
     s = sha_new()
-    s.update(message)
+    s.update(message[:SCRAMBLE_LENGTH]) 
     s.update(stage2)
     result = s.digest()
     return _my_crypt(result, stage1)


elemount:
This fix is related with issue #584 . This is the auth switch error. 
If you like, you can connect following database and use wireshark to watch the procedure.
```python
conn = pymysql.connect(host='shuodlclient.mysql.database.azure.com', port=3306, user='mysqlaas@shuodlclient', passwd='test_PyMySQL', db='mysql')
```

elemount:
And this is also captured by mysqljs/mysql#1730 .  @bgrainger notice that.

methane:
Do you mean

* Azure using buggy MySQL server which sends wrong packet?
* Azure requires old_password which is deprecated and not secure?
* All MySQL client library should do such "wrong" fix?

methane:
I'm sorry, `_scramble` is not old_password.

elemount:
```
Azure using buggy MySQL server which sends wrong packet?
```
No, Azure just send the packet which MySQL server sent on auth switch stage.
```
Azure requires old_password which is deprecated and not secure?
```
No, do not use old_password
```
All MySQL client library should do such "wrong" fix for Azure?
```
Not for azure, it is the fix for MySQL. Our PyMySQL do not even response correct AuthSwitch packet. All the default drivers like libmysqlclient and libmariadbclient will response correctly. A node-mysql2 and go-sql-driver/mysql also response correctly. 

I provide a Azure MySQL server for test because it is easy to repo by it. I can not just provide my local MySQL for you to repo it.

elemount:
Both MySQL and MariaDB has this document and behavior mismatch. 
From code, for example 
For current version 10.2 of MariaDB, all the auth code is in sq_acl.cc. Change Plugin Packet function send_plugin_request_packet is called by function server_mpvio_write_packet and this function is called by a server authentication plugin, when it wants to send data to the client. All this function are just pass the data of the plugins. native_password_authenticate and old_password_authenticate are both use code looks like

mpvio->write_packet(mpvio, (uchar*)thd->scramble, SCRAMBLE_LENGTH + 1)

So the response will contains string[NUL].

elemount:
I have a way to repo.
1. Create publick_key.pem and private_key.pem by openssl
```
openssl genrsa -out private_key.pem 2048
openssl rsa -in private_key.pem -pubout -out public_key.pem
chmod 400 private_key.pem
chmod 444 public_key.pem
```
2. Install MySQL from source. And initialize it. Add following line to the `my.cnf` of base folder.
```
sha256_password_private_key_path = private_key.pem
sha256_password_public_key_path = public_key.pem
default-authentication-plugin=sha256_password
``` 
3. Start mysql, and login it. create a user 'nativeuser'@'localhost' by 
```sql
CREATE USER 'nativeuser'@'localhost' IDENTIFIED WITH mysql_native_password;
SET old_passwords = 0;
SET PASSWORD FOR 'nativeuser'@'localhost' = PASSWORD('test');
GRANT ALL PRIVILEGES ON *.* TO 'nativeuser'@'localhost' WITH GRANT OPTION;
FLUSH PRIVILEGES
```
4. Test the nativeuser with the test password. 
And notice, you may need use the SHA256 pr as the pymysql code.

methane:
I don't want to pay my time for test it.
Could you create test using docker on Travis?

elemount:
You can notice old password encrypt use the scream length because it hint the error by auth switch. Or you can remove the length split of old password and run the tests.

methane:
It's not good evidence, because two are different.
But I confirmed mysql-connector-c's client code use constant SCRAMBLE_LENGTH.

PR #591, opened 2017-08-21
new charset definition added

peakkk:
When connecting to MySQL 8, PyMySQL may raise an exception, such as `KeyError: 255`.
To solve the problem, serval new charset definition was added to this file.


methane:
See #577

peakkk:
There is the chareset definition from my brand-new MySQL 8 server.
charset

And when I user PyMySQL to connect to the server, an exception about charset raised.
exception

methane:
Doesn't they conflict with other servers compatible MySQL protocol, especially MariaDB?
Is there any reason to add collation other than 255?
Especially, collation with id>255 can't be used in protocol.

methane:
Since MySQL 8 is not GA yet, I don't want to add them now
even if they doesn't conflict with MariaDB.

timloo0710:
I encountered the same problem, 

How did I add 46 character set ? Where can I get the information of the character set 46? 

My mysql server is stated on the Windows Server 2012 R2 . 

Below is my error message!! Please Ref.

methane:
What version of MySQL do you use?
Why don't you use utf-8?

timloo0710:
default

I can't figure out what I setup the language encode option. 
Where can I modify the option??

Thanks!

methane:
https://github.com/PyMySQL/PyMySQL#example

Why don't you read README before asking question???

https://medium.com/@methane/why-you-must-not-ask-questions-on-github-issues-51d741d83fde#.uoo6bjv13

GOODLIFEZW:
it solves my problem， thank you very much

PR #593, opened 2017-08-23
Fields with binary data will now have their converters set to 'None'.

CodePanter:
Previously binary data would be incorrectly converted. This changeset prevents this from happening by setting the converter to 'None' when the field.charsetnr equals 63, regardless of whether Unicode is being used. Everything else is still treated the same.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -141,7 +141,7 @@ def _scramble(password, message):
     stage1 = sha_new(password).digest()
     stage2 = sha_new(stage1).digest()
     s = sha_new()
-    s.update(message[:SCRAMBLE_LENGTH]) 
+    s.update(message[:SCRAMBLE_LENGTH])
     s.update(stage2)
     result = s.digest()
     return _my_crypt(result, stage1)
@@ -807,7 +807,7 @@ class Connection(object):
 
     def escape(self, obj, mapping=None):
         """Escape whatever value you pass to it.
-        
+
         Non-standard, for internal use; do not use this in your applications.
         """
         if isinstance(obj, str_type):
@@ -816,7 +816,7 @@ class Connection(object):
 
     def literal(self, obj):
         """Alias for escape()
-        
+
         Non-standard, for internal use; do not use this in your applications.
         """
         return self.escape(obj, self.encoders)
@@ -1487,6 +1487,7 @@ class MySQLResult(object):
         description = []
 
         for i in range_type(self.field_count):
+            isbinary = False
             field = self.connection._read_packet(FieldDescriptorPacket)
             self.fields.append(field)
             description.append(field.description())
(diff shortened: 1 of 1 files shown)


methane:
example and test?

CodePanter:
Are there specific kinds of tests and examples you would like me to use to illustrate and support this pull request?

methane:
Test case failing without this fix.

methane:
And I can't understand what pull request title means.

methane:
Additionally, all added `if` statement looks ugly.
I believe your fix should be implemented more simply.

JanKlopper:
Your original code sets the encoding to None for binary data fields, which is correct.
However, it still converts them into text for both unicode connections and non unicode connections which isn't very useful, as they are binary fields, not text fields.

The lookup on line 1513 cannot work without an extra variable (isbinary) tracking if the field was binary since its field_type will not be enough to differentiate from actual text file types.

pymysql/connections.py
```diff
@@ -1498,19 +1499,24 @@ def _get_descriptions(self):
...
+
+            if field_type in TEXT_TYPES and field.charsetnr == 63:  # binary
+                # TEXTs with charset=binary means BINARY types.
+                encoding = None
+                isbinary = True
+
+            if not isbinary:
+                converter = self.connection.decoders.get(field_type)
+            else:
+                converter = None
```
methane: Why isbinary needed?  Why is this wrong?
Anyway, test for use_unicode=False is needed.

```py
            if field_type in TEXT_TYPES and field.charsetnr == 63:  # binary
                # TEXTs with charset=binary means BINARY types.
                encoding = None
                converter = None
            else:
                converter = self.connection.decoders.get(field_type)
```

JanKlopper:
Sure that works.

pymysql/connections.py
```diff
@@ -1498,19 +1499,24 @@ def _get_descriptions(self):
...
-                    encoding = conn_encoding  # SELECT CAST(... AS JSON) 
-                elif field_type in TEXT_TYPES:
-                    if field.charsetnr == 63:  # binary
-                        # TEXTs with charset=binary means BINARY types.
-                        encoding = None
-                    else:
-                        encoding = conn_encoding
+                    encoding = conn_encoding  # SELECT CAST(... AS JSON)
+                elif field_type in TEXT_TYPES and field.charsetnr != 63:  # binary
+                    encoding = conn_encoding
```
methane: This is ugly too.  Since encoding will be overridden, `elif field_type in TEXT_TYPES:` will be enough.
But I believe there must be more simpler code.

methane:
> However, it still converts them into text for both unicode connections and non unicode connections which isn't very useful, as they are binary fields, not text fields.

I can't understand it.  I admit non unicode connection is not well tested.  (I never use the option.
contribution is welcome).
But unicode connection is tested, and we can get binary data without converting it to text.

https://github.com/PyMySQL/PyMySQL/blob/cee594543821f464d94c5962c23b3d11f7900899/pymysql/tests/test_basic.py#L94-L104

Am I wrong?

JanKlopper:
Binary fields are different from blob fields.
````
    def test_binary(self):
        """test binary data"""
        data = bytes(bytearray(range(256)) * 4)
        conn = self.connections[0]
        self.safe_create_table(
            conn, "test_binary", "create table test_binary (b binary)")

        with conn.cursor() as c:
            c.execute("insert into test_binary (b) values (%s)", (data,))
            c.execute("select b from test_binary")
            self.assertEqual(data, c.fetchone()[0])
````

**Comment by methane** [comment]
Do you mean your "test_binary" is not in current PyMySQL test cases?
Could you add it?

**Comment by methane** [comment]
I still don't understand what's wrong with you.
Could you explain by user code?

File: pymysql/connections.py (hunk 2 above)
Comment (methane) [reply pymysql/connections.py]: This is more clean.

```py
if field_type in TEXT_TYPES and field.charsetnr == 63:  # binary
    encoding = converter = None
else:
    converter = self.connection.decoders.get(field_type)
    if field_type in TEXT_TYPES or field_type == FIELD_TYPE.JSON:
        encoding = conn_encoding if use_unicode else None:
    else:
        encoding = 'ascii'
```

methane: Oh, wait.  Let's skip converters for text types too.

```py
converter = None
if field_type in TEXT_TYPES and field.charsetnr == 63:  # binary
    encoding = None
elif field_type in TEXT_TYPES or field_type == FIELD_TYPE.JSON:
    encoding = conn_encoding if use_unicode else None
else:
    encoding = 'ascii' if use_unicode else None
    converter = self.connection.decoders.get(field_type)
```

methane:
I fixed #596, and it passes.

PR #596, opened 2017-08-30
Add binary field type test to basic tests, currently fails where blob works

JanKlopper:


pymysql/tests/test_basic.py
```diff
@@ -91,11 +91,24 @@ def test_integer(self):
...
         """test binary data"""
         data = bytes(bytearray(range(256)) * 4)
         conn = self.connections[0]
         self.safe_create_table(
+            conn, "test_binary", "create table test_binary (b binary)")
+
+        with conn.cursor() as c:
+            c.execute("insert into test_binary (b) values (%s)", (data,))
+            c.execute("select b from test_binary")
+            print data, c.fetchone()[0]
```
methane: remove it.  Travis fails because of this.

pymysql/tests/test_basic.py
```diff
@@ -91,11 +91,24 @@ def test_integer(self):
...
-    def test_blob(self):
+    def test_binary(self):
         """test binary data"""
         data = bytes(bytearray(range(256)) * 4)
         conn = self.connections[0]
         self.safe_create_table(
+            conn, "test_binary", "create table test_binary (b binary)")
+
+        with conn.cursor() as c:
+            c.execute("insert into test_binary (b) values (%s)", (data,))
```
methane: Travis test with Python 2.7 fails by "Data truncated for column 'b' at row 1" warning.

JanKlopper:
Sorry about that.
````
    def test_binary(self):
        """test binary data"""
        data = bytes(bytearray(range(255)))
        conn = self.connections[0]
        self.safe_create_table(
            conn, "test_binary", "create table test_binary (b binary(255))")

        with conn.cursor() as c:
            c.execute("insert into test_binary (b) values (%s)", (data,))
            c.execute("select b from test_binary")
            self.assertEqual(data, c.fetchone()[0])
````
This seems to be valid and pass the test basic_test.
However, as the old code seems to str() call that data from the database for binary fields, and unicode cast that data from the database it triggers a UnicodeDecodeError if there are any non encodeable chars in the binary data.

Depending on the data in in the binry field this leads to an error like:
`UnicodeDecodeError: 'ascii' codec can't decode byte 0xe0 in position 4: ordinal not in range(128)` since the decoders will think its a text field.

methane:
Python 2? 3?
Anyway, test failing with current code is required.

PR #597, opened 2017-08-30
allow no whitespaces between `VALUES` and `(` when do bulk insert/replace

uwydoc:
the [regular expression](../blob/master/pymysql/cursors.py#L15) used to check whether bulk insert/replace is possible is as follows:
```python
RE_INSERT_VALUES = re.compile(
    r"\s*((?:INSERT|REPLACE)\s.+\sVALUES?\s+)" +
    r"(\(\s*(?:%s|%\(.+\)s)\s*(?:,\s*(?:%s|%\(.+\)s)\s*)*\))" +
    r"(\s*(?:ON DUPLICATE.*)?);?\s*\Z",
    re.IGNORECASE | re.DOTALL)
```
notice the last `\s+` in the first part of the regular expression requires whitespaces between `VALUES` and left-parenthesis, which means queries like `insert into user values(%s, %s)` would not be considered bulk-insertable. although the demo sql query in the comment of `executemany` has a space between `VALUES` and `(`, i do think this restriction could be relaxed.

This gotcha is so trivial and would cost some time to realize. It actually costed me some time to find out why the performance of a relatively simple program dropped by more than 6 times after switching from MySQLdb to PyMySQL.

The fix is quite simple, just replace the `\s+` with `\s*` in the first part.

**NOTE:** ported from https://github.com/PyMySQL/mysqlclient-python/commit/354dcb59e286b0db3b108c92760d3ec135a8bc23

### DIFF
--- a/pymysql/cursors.py
+++ b/pymysql/cursors.py
@@ -12,7 +12,7 @@ from . import err
 #: executemany only suports simple bulk insert.
 #: You can use it to load large dataset.
 RE_INSERT_VALUES = re.compile(
-    r"\s*((?:INSERT|REPLACE)\s.+\sVALUES?\s*)" +
+    r"\s*((?:INSERT|REPLACE)\b.+\bVALUES?\s*)" +
     r"(\(\s*(?:%s|%\(.+\)s)\s*(?:,\s*(?:%s|%\(.+\)s)\s*)*\))" +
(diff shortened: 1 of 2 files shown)


methane:
Please copy this commit:
https://github.com/PyMySQL/mysqlclient-python/commit/354dcb59e286b0db3b108c92760d3ec135a8bc23

uwydoc:
@methane done and thanks.

PR #598, opened 2017-08-30
(proposal) Add json decoder

fpietka:
This pull request is a proposal to add JSON decoder.

I don't know if this is change make sense for every user, hence the proposal.

### DIFF
--- a/pymysql/converters.py
+++ b/pymysql/converters.py
@@ -4,6 +4,7 @@ import datetime
 from decimal import Decimal
 import re
 import time
+import json
 
 from .constants import FIELD_TYPE, FLAG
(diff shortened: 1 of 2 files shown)


methane:
No. Because:

* Backward incompatible
* You can't pass non serialized data to execute (e.g. `cur.execute("SELECT ?", ({"foo", "bar"},))` is inpossible.)  So it's not symmetrical.

You can decode json after fetching, not while fetching.
It doesn't have any performance overhead.
When error happen while decoding, traceback from deep inside PyMySQL is not readable.
So decode manually after fetch makes your code more maintainable.

PR #609, opened 2017-10-08
remove shebangs and unusable main code

meeuw:
remove shebangs and unusable main code

these python scripts don't contain "main code" (to be executed stand only), I don't see why a shebang (#!) should be added.

the removed "main code" is unusable (as far as I know), when executing it gives the following error: `ImportError: cannot import name 'capabilities'`

The shebang in these scripts are against the packing guidelines for Fedora and requires this patchwork:
http://pkgs.fedoraproject.org/cgit/rpms/python-PyMySQL.git/tree/python-PyMySQL.spec#n64


methane:
thanks

PR #613, opened 2017-10-20
add string cast for database, user and password to correctly manage

the4thdoctor:
A pg_chameleon user noticed that using numerical passwords can cause issues because the password could be wrongly interpreted as an integer. Here the details of the problem https://github.com/the4thdoctor/pg_chameleon/issues/31

I'm fixing the issue on my side. This PR should fix the problem on the source :)

Thanks

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -652,9 +652,9 @@ class Connection(object):
 
         self.host = host or "localhost"
         self.port = port or 3306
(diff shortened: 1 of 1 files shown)


methane:
I don't like it. Password must be string, not int.

the4thdoctor:
In that case you should file a feature request to mysql which allows numerical users and passwords.
no problem. I've already fixed the problem on my side.

PR #615, opened 2017-10-31
convert float value 'nan' to NULL

wanglun:


methane:
"Errors should never pass silently." -- https://www.python.org/dev/peps/pep-0020/

If you want to convert nan to NULL, do it explicitly.
I don't want do it in PyMySQL implicitly.

PR #621, opened 2017-11-23
more short import in example

juanpabloaj:


methane:
The example uses `pymysql.cursors.DictCursor`.
There are no guarantee `pymysql.cursor` is imported only by `import pymysql`.

While there are no plan for such change, longer explicit import is preferred,
especially in official example.

PR #625, opened 2017-12-14
add syn /async connection pools

ilqxejraha:
I have add connection pool that support syn and async.

syn pool , I have use it to my production env.
async pool, there's something wrong with tornado 4.5, I hope someone could help me to fix it.


methane:
I don't want to increase code size to maintain.
You should release it by yourself, as separated package.

ilqxejraha:
Thanks for your review. I will delete it.
But there's nothing connection pool for pymysql, for example, tormysql , it based on tornado

I hope someone could support connection pool module for PyMySQL.

PR #633, opened 2018-01-15
add AttrDictCursor

caitinggui:
Let results allow attribute-style access, and I think it can be a bit user-friendly.
data = {"row1": "the first row", "row2": "foo"}
data["row1"]   VS  data.row1

### DIFF
--- a/pymysql/cursors.py
+++ b/pymysql/cursors.py
@@ -396,6 +396,24 @@ class DictCursor(DictCursorMixin, Cursor):
     """A cursor which returns results as a dictionary"""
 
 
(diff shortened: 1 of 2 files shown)


methane:
I don't like it

caitinggui:
But it can be more convenient and without any performance loss.

methane:
You can use it even if PyMySQL doesn't provide it.
And why do you think your AttrDict is world best?  Is yours better than attrdict on PyPI?

Anyway, I want to focus low-level, performance and stability.
I don't want to add user-friendly parts I maintain.

PR #634, opened 2018-01-16
CHANGELOG - fix note about MULTI_STATEMENTS

adamchainz:
Improving the grammar and correcting the flag name to `MULTI_STATEMENTS` (with an 's').


methane:
thanks

Issue #297, opened 2015-02-02
Add speedup module for CPython


methane:
There are two bottlenecks.

1. Reading packet
2. Parsing packet

1 is resolved by using io module.

To resolve issue 2, we should remove classes for parsing packets, and use simple flat function and
builtin types.  It make PyMySQL on PyPy faster.
Then, we can easily replace them with speedup module written in C.

zcattacz:
I have script periodically pump excel data into mariadb using pymysql.
each time it needs to send ~50k records. it takes about 5.5 min.
After compiling pymysql module with Nuitka, the time is lowered to 3.2min.
I consider it a good speed gain.

groutr:
What is the status of this speedup module?

horacekmi:
I was able to speed up reading/parsing packets by roughly 28% just by reducing the number of function calls. Reading ~70k rows (x45 cols) dropped from 8s to 5.7s on my machine. This quick edit however duplicates some code. See my attempt here: https://github.com/horacekmi/PyMySQL/commit/14125ef1026553c88e28ef3d7bdb849676edd102

byaka:
micro-optimisations, that converts source from hierarchical to linear really can burst performance near 20% in some cases, but it bad way. much better to rewrite thats parts of code to cython...
anyway i agree with TS, but who has time for full rewrite of parser? :(

groutr:
I'd be willing to make an initial attempt at isolating and simplifying the parsing code (issue 2 from above).

rspadim:
i was using aiomysql, and found a point to speed up

when i was executing a insert ignore, or something that could report a known warning, it was executing a show warnings (2 queries)

force removing the show warnings changed my app to run from ~7kqps to ~10kqps (multi process)

i didn't checked pymysql code, but this could be a nice parameter at cursor.execute() function

rspadim:
and maybe we could use numba+cython, when possible?

kesmit13:
I opened another similar issue (#1066) not realizing this one was here. I have a fork of PyMySQL with a C extension that replaces the row data reader method and adds some tuned cursors that gives significant speed improvements. The issue was opened to start a conversation about a possible plugin for PyMySQL, and to collaborate on further improvements.

PR #638, opened 2018-01-23
[WIP] Refactor parsing code

groutr:
This PR is a still a work in progress.  However, it is in a state that people can start looking at it.  It passes most of the PyMySQL test suite.  The errors I'm getting from the test suite could be caused by testing against MariaDB 10.1.22 (mostly SQL syntax errors).

This PR addresses the second point suggested in https://github.com/PyMySQL/PyMySQL/issues/297 by @methane 

> To resolve issue 2, we should remove classes for parsing packets, and use simple flat function and
builtin types. It make PyMySQL on PyPy faster.
Then, we can easily replace them with speedup module written in C.

Idea behind the new parsing code.  Separate data from the operations on the data.  MysqlPackets are now represented using namedtuples, though switching to plain tuples may provide a small performance increase.  Instead of wrapping generic response packets in custom objects, I simply return a tuples or dictionaries.

The parsing functions try to be very simple and not have state.  There are three classes of parsing functions: 

1. Data parsers: These functions are the lowest level parsing functions.  They accept a data argument and an offset keyword argument.  Their return type is a tuple, (# of bytes read, data read). The first value of the tuple is used for tracking progress through the data (I figured it was better than slicing data or creating memoryviews).  The second value of the tuple is the data decoded by that parsing function.
```python
def read_uint8(data, offset=0):
    return 1, data[offset]    # (size of uint8, value of uint8)
```
2. Packet parsers: The next layer of functions is functions for parsing whole packets.  These functions accept a Packet namedtuple and potentially other arguments.  Examples are `parse_ok_packet` and `parse_field_descriptor_packet`.  These packet parsing functions return a dictionary of results.
3. Stream parsers: The highest layer are functions that take a stream of packets along with potentially any keyword arguments needed to parse the stream.  The stream is an iterable of packets such as a result set.  There is no set return pattern for these functions.  Example: `parse_result_stream`.  The return is a generator that first yields field information, then rows as the are parsed.

I think this layout is conducive to a faster Python implementation and ultimately a Cython implementation.  I haven't started profiling and fine tuning yet as I wanted to just get the things in place and working first.
I will appreciate any feedback.


groutr:
One idea I have for a possible performance improvement is caching Python objects from rows.  If rows contain a lot of the same values (one column is constant or a small range of values), we can use an already constructed Python object (keyed by (field, bytes)).  This should reduce the number of Python objects that have to be created and could be beneficial for large result sets.

groutr:
I've been experimenting with a Cythonized version of this parser.  I'm using the benchmark found here to provide some very quick numbers.
```
CPython (master):
6.947011
6.970348
6.859433

PyPy3 (master):
1.968035
1.978064
1.946565


CPython (parser2):
6.754084
6.500964
6.593116

PyPy3 (parser2):
2.080422
2.031008
2.030981

CPython (parser2 + Cython):
4.588248
4.586170
4.609609
```
There is still some more optimization to be done in the reading packets.  Hopefully I'll have my ideas down into by the end of this week.

methane:
Please don't try Cython in this pull request.

methane:
@groutr Is this PR ready to review?

I don't want to use Cython.  I want to use C instead, via normal extension module or CFFI.
But it can be done in other pull request.

groutr:
@methane I'll remove the cython.  It was mostly there just to see what sort of performance gains could be had. I'd appreciate feedback.  I have some other improvements that I need to push still that should give some more performance, but need to benchmark/test first.

groutr:
@methane Can I get your thoughts on using generators in the parser?  Specifically on the _iter_result_packets and parse_result_stream functions.

methane:
Sorry, this pull request is too large to review.

> Specifically on the _iter_result_packets and parse_result_stream functions.

It looks ugly.

methane:
Would you try step-by-step approach, instead of very big pull request?

* Move low-level codes to `pymysq.protocol` module, without changing anything.
* Single refactoring * many pull requests

All pull requests should passes test.

methane:
I prepared `devel` branch for it.

groutr:
@methane I agree this PR has gotten very messy.  This branch represents several iterations/attempts, experimenting with what seemed to work better and what didn't.  I like the devel branch approach over a single PR.

PR #639, opened 2018-01-28
Auto ping

toien:
This is to avoid connection has been idle for long time and mysql server has closed it.

When try to execute query on this connection again, `err.InterfaceError("(0, '')")` or `pymysql.err.OperationalError: (2013, 'Lost connection to MySQL server during query')` raised.

After set `auto_ping` param on `Connection` construct function, every time connection do query, it will `ping()` first.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -576,7 +576,7 @@ class Connection(object):
                  autocommit=False, db=None, passwd=None, local_infile=False,
                  max_allowed_packet=16*1024*1024, defer_connect=False,
                  auth_plugin_map={}, read_timeout=None, write_timeout=None,
-                 bind_address=None, binary_prefix=False):
+                 bind_address=None, binary_prefix=False, auto_ping=False):
         if no_delay is not None:
             warnings.warn("no_delay option is deprecated", DeprecationWarning)
 
@@ -681,6 +681,8 @@ class Connection(object):
 
         #: specified autocommit mode. None means use server default.
         self.autocommit_mode = autocommit
+
+        self.auto_ping = auto_ping
 
         if conv is None:
             conv = converters.conversions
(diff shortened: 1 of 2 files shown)


methane:
I don't like it.
You or ORM can do it manually.

toien:
Thanks for fast reply!

This feature is configurable and useful to long-time running app. Because of python's GIL, i think a lot python programs will have only one connection per process.

`auto_ping` makes reconnect transparent to client code, letting user use PyMySQL more easily, no need to care for details.

By now, i have use this code on product env, and i seems work good. 

Hope you could reconsider it, which doing a big help on long-time app while change little source code.

methane:
> This feature is configurable

Configurable is not problem.  I want to minimize options and features I maintain.

> useful to long-time running app. 

Usefulness is not my primary target.  Small, simple, robust, stable and performance is my target.
In other words, if the feature can be implemented on top of this library, it's not my goal.
If the feature should be implemented in lowest level, (e.g. configure some TCP options)
I'll consider it.

> auto_ping makes reconnect transparent to client code, letting user use PyMySQL more easily, no need to care for details.

I don't think this is best practice.  I prefer some other approaches and all of them are implemented
on top of driver.

These are example of some approaches I prefer.  All of them are implemented in pool layer and they are driver independent.

* My favorite is Go's [SetConnMaxLifetime(]https://golang.org/pkg/database/sql/#DB.SetConnMaxLifetime).  (I implemented this feature.)
* auto ping when check out connection from pool.  [SQLAlchemy have it](http://docs.sqlalchemy.org/en/latest/core/pooling.html#sqlalchemy.pool.Pool.params.pre_ping)

If you like your approach and make it reusable, I recommend to publish it on PyPI and maintain by yourself.
Please don't bother me.

PR #641, opened 2018-02-16
Change password to passwd for making connection

opensubb009:
return Connection(*args, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'password'
Due to this exception, need to change password to passwd for making connection.


methane:
https://github.com/PyMySQL/PyMySQL/blob/c51b47e3c6fa0d8b8090a6b25f7168ed68082dea/pymysql/connections.py#L569

PR #642, opened 2018-02-24
Improve documentation with errors and links

Phlosioneer:
I added several links to the PEP 249 documentation. I also documented
the various exceptions that can be raised and, clarified the meaning
of some parameters.

Overall, readers will be able to cross-reference methods with the
specification, and will see implementation-specific info. This will
make it clear that this library implements the full specification,
which is important because there is almost no direct documentation
of this library.

I've confirmed that the docs build correctly using the makefile in `docs/`.

Let me know if I missed anything, or if there are any style issues.


Phlosioneer:
Note: Build errors don't seem to be related to my code.

methane:
Thanks.

PR #648, opened 2018-04-15
Make it possible to actually unblock a blocking receive

Photonios:
Packages such as https://github.com/noplay/python-mysql-replication call `_read_packet`, which blocks forever. Evening closing the underlying connection with `.close()` causes nothing to happen.

Calling `socket.shutdown` on the socket actually causes the blocking read to get unstuck. On top of that, if a read returns `EINTR`, you're not supposed to just blindly continue. You're supposed to check if you might need to unblock. Even though this never happens because the socket is in blocking mode.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -1065,6 +1065,8 @@ class Connection(object):
                 break
             except (IOError, OSError) as e:
                 if e.errno == errno.EINTR:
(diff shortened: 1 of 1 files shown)


methane:
Could you provide demo code illustrate socket.close cause EINTER?

PR #652, opened 2018-04-28
test_plugin: only for MySQL-5.5 and fix plugin assumptions (which was wrong in MySQL-8.0)

grooverdan:


methane:
I feel current code is ugly and fragile.  But it's OK for now...

PR #664, opened 2018-04-29
test_set_charset: use utf8mb4

grooverdan:


methane:
MySQL 5.5 supports utf8mb4.
Since we dropped MySQL 5.1, how about utf8mb4 always?

grooverdan:
> how about utf8mb4 always?

Sure.

PR #669, opened 2018-05-08
Protocol module

groutr:
Move parsing functions to a protocol module.

There are three main types of parsing functions: packet-level, fixed length, and variable length.  Packet-level functions operate on the whole packet and do not accept an offset parameter.  Fixed-length functions parse a known amount of bytes and return the only the result.  Variable-length functions parse a variable amount of data and return a tuple of (bytes read, result).

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -23,6 +23,7 @@ from .cursors import Cursor
 from .optionfile import Parser
 from .util import byte2int, int2byte
 from . import err
+from . import protocol
 
 try:
     import ssl
@@ -276,34 +277,28 @@ class MysqlPacket(object):
         """
         return self._data[position:(position+length)]
 
-    if PY2:
-        def read_uint8(self):
-            result = ord(self._data[self._position])
-            self._position += 1
-            return result
-    else:
-        def read_uint8(self):
-            result = self._data[self._position]
-            self._position += 1
-            return result
+    def read_uint8(self):
+        result = protocol.read_uint8(self._data, offset=self._position)
+        self._position += 1
+        return result
 
     def read_uint16(self):
-        result = struct.unpack_from('<H', self._data, self._position)[0]
+        result = protocol.read_uint16(self._data, offset=self._position)
         self._position += 2
         return result
 
     def read_uint24(self):
-        low, high = struct.unpack_from('<HB', self._data, self._position)
+        result = protocol.read_uint24(self._data, offset=self._position)
         self._position += 3
-        return low + (high << 16)
+        return result
 
     def read_uint32(self):
-        result = struct.unpack_from('<I', self._data, self._position)[0]
+        result = protocol.read_uint32(self._data, offset=self._position)
         self._position += 4
         return result
 
     def read_uint64(self):
-        result = struct.unpack_from('<Q', self._data, self._position)[0]
+        result = protocol.read_uint64(self._data, offset=self._position)
         self._position += 8
         return result
 
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -296,7 +296,6 @@ class MysqlPacket(object):
 
     def read_string(self):
         pos, result = protocol.read_string(self._data, offset=self._position)
-        # We need to add one to account for the null terminating character.
         self._position += pos
         return result
 
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -317,10 +317,9 @@ class MysqlPacket(object):
         (unsigned, positive) integer represented in 1-9 bytes followed by
         that many bytes of binary data.  (For example "cat" would be "3cat".)
         """
-        length = self.read_length_encoded_integer()
-        if length is None:
-            return None
-        return self.read(length)
+        bytes_read, result = protocol.read_length_coded_string(self._data, offset=self._position)
+        self._position += bytes_read
+        return result
 
     def read_struct(self, fmt):
         s = struct.Struct(fmt)
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -303,11 +303,9 @@ class MysqlPacket(object):
         return result
 
     def read_string(self):
-        end_pos = self._data.find(b'\0', self._position)
-        if end_pos < 0:
-            return None
-        result = self._data[self._position:end_pos]
-        self._position = end_pos + 1
+        pos, result = protocol.read_string(self._data, offset=self._position)
+        # We need to add one to account for the null terminating character.
+        self._position += pos
         return result
 
     def read_length_encoded_integer(self):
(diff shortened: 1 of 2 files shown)


pymysql/connections.py
```diff
@@ -276,43 +269,34 @@ def get_bytes(self, position, length=1):
...
-            return result
-    else:
-        def read_uint8(self):
-            result = self._data[self._position]
-            self._position += 1
-            return result
+    def read_uint8(self):
+        result = protocol.read_uint8(self._data, offset=self._position)
+        self._position += 1
+        return result
```
methane: Don't split methods.  Just move.

methane:
I did "just move".  What do you want to do after that?

groutr:
@methane I was under the impression that the final goal was to remove the wrapping classes and use calls to simple functions with builtin datatypes. That's why I moved the actual parsing code into their own functions, because eventually the wrapping classes would go away.

From https://github.com/PyMySQL/PyMySQL/issues/297#issue-56198431
> To resolve issue 2, we should remove classes for parsing packets, and use simple flat function and
> builtin types.

Phase 2 would be separating out the functions that orchestrate packet parsing (like _read_ok_packet, _read_rowdata_packet, etc).  In this PR, I attempted to remove half of the work of MysqlPacket to parsing functions.  The next phase would be to remove the other half (the position tracking inside packets) to functions that orchestrate the overall process of parsing packets.  At that point MysqlPacket isn't doing anything and can be removed along with its companion classes.

The next phase would be replacing MysqlPacket with a tuple. This means that _read_packet would be returning tuples instead of instances of classes.
Some formats I've been considering are:
1. `(packet number, packet data)`: This would probably be the clearest replacement. This could allow for interesting packet parsing strategies, such as processing packets in parallel.  
2. `(packet data,)`: Most of the time, all that is actually needed is packet data.  After the packet number is checked by _read_packet, it doesn't seem to be used again.
3. `(packet number, packet length, packet data)`: This is the representation I used in my first attempt.  It worked well, except packet length wasn't used all much, and is trivial to get by calling `len(packet data)`.
4. `(packet number, packet type, packet data)`: Similar to 1, but we also store the packet type.  This would ensure that the packet type is determined exactly once. I'm unclear on the precise benefits though.

At that point, I believe the goal set forth initially will have been fulfilled. A C extension can then be created to significantly accelerate packet parsing and the orchestrating functions can use either the C-extension or the pure-python module.

groutr:
@methane, any feedback on the above comment?

methane:
> I was under the impression that the final goal was to remove the wrapping classes and use calls to simple functions with builtin datatypes.

OK,

> That's why I moved the actual parsing code into their own functions, because eventually the wrapping classes would go away.
> ...
> In this PR, I attempted to remove half of the work of MysqlPacket to parsing functions.

I felt this was wrong step.  It only doubles the function call (= performance), without any simplification.

> The next phase would be to remove the other half (the position tracking inside packets) to functions that orchestrate the overall process of parsing packets. At that point MysqlPacket isn't doing anything and can be removed along with its companion classes.

I think next step should convert PacketWrapper classes into simple function.
It can be achieved without touching MySQLPacket class.
It would be easy to review, and no performance penalty.

Maybe, FieldDescriptorPacket can be converted too.

Breaking MySQLPacket class will be next step, but I haven't decided to do it or not.

groutr:
> 
>     That's why I moved the actual parsing code into their own functions, because eventually the wrapping classes would go away.
>     ...
>     In this PR, I attempted to remove half of the work of MysqlPacket to parsing functions.
> 
> I felt this was wrong step. It only doubles the function call (= performance), without any simplification.

This is an intermediate step, so temporarily increasing function call wasn't a concern.  It was done to show that no functionality was being broken by using the simple functions.  It meant that any call to MySqlPacket.read_uint8 could be replaced with a call to protocol/read_uint8 as they were now equivalent.

>     The next phase would be to remove the other half (the position tracking inside packets) to functions that orchestrate the overall process of parsing packets. At that point MysqlPacket isn't doing anything and can be removed along with its companion classes.
> 
> I think next step should convert PacketWrapper classes into simple function.
> It can be achieved without touching MySQLPacket class.
> It would be easy to review, and no performance penalty.

I feel at this stage, performance is the wrong goal.  We should be focusing on getting the implementation correct for an easy C extension, and then optimizing that implementation.  The endgame is to prepare things for a C extension which is where performance will become a primary goal.

> Maybe, FieldDescriptorPacket can be converted too.
> 
> Breaking MySQLPacket class will be next step, but I haven't decided to do it or not.

Again, I was under the impression that removing MySQLPacket and the other wrapping classes was the whole point of this effort.  I don't understand your approach very well. Could you outline your steps more completely?

methane:
2018年6月6日(水) 23:30 Ryan Grout <notifications@github.com>:

> That's why I moved the actual parsing code into their own functions, because eventually the wrapping classes would go away.
> ...
> In this PR, I attempted to remove half of the work of MysqlPacket to parsing functions.
>
> I felt this was wrong step. It only doubles the function call (=
> performance), without any simplification.
>
> This is an intermediate step, so temporarily increasing function call
> wasn't a concern. It was done to show that no functionality was being
> broken by using the simple functions. It meant that any call to
> MySqlPacket.read_uint8 could be replaced with a call to protocol/read_uint8
> as they were now equivalent.
>

I want defer degrading medium step as possible.  I may make a release
before final step.

The next phase would be to remove the other half (the position
tracking inside packets) to functions that orchestrate the overall
process of parsing packets. At that point MysqlPacket isn't doing
anything and can be removed along with its companion classes.
>
> I think next step should convert PacketWrapper classes into simple
> function.
> It can be achieved without touching MySQLPacket class.
> It would be easy to review, and no performance penalty.
>
> I feel at this stage, performance is the wrong goal. We should be focusing
> on getting the implementation correct for an easy C extension, and then
> optimizing that implementation. The endgame is to prepare things for a C
> extension which is where performance will become a primary goal.
>

C extension is optional, and may be not used on PyPy.  And I may make
release before C extension.
So pure Python performance is important.

Maybe, FieldDescriptorPacket can be converted too.
>
> Breaking MySQLPacket class will be next step, but I haven't decided to do
> it or not.
>
> Again, I was under the impression that removing MySQLPacket and the other
> wrapping classes was the whole point of this effort. I don't understand
> your approach very well. Could you outline your steps more completely?
>

1. Reduce unnecessary wrapper classes.
2. Maybe, make a release, with MySQL 8 support.
3. Survey tools relating to this project. (E.g. mycli, aiomysql, maybe
more.)
4. Think what should be done.

> --
INADA Naoki  <songofacandy@gmail.com>

groutr:
Isn't the point of this branch to isolate the work until it's ready to merge to master?  Releases can still be made from master, independently of the state of this branch.

methane:
I don't make decision yet.

But master branch is only for bugfix.
I'll implement any new feature in devel branch.

2018年6月7日(木) 14:54 Ryan Grout <notifications@github.com>:

PR #680, opened 2018-05-29
Remove unused `fields` from result

methane:


methane:
`field` is used... It would be converted into namedtuple.

PR #682, opened 2018-06-08
Add sha256 and chaching sha2 auth support

methane:
This pull request is continued from #583


methane:
> I hoped to reuse it in aiomysql

For now, supporting MySQL 8 is higher priority than providing stable API.

But it's OK to reuse it with risk of future breakage.
Aiomysql reuse many private code already.

PR #684, opened 2018-06-12
Make `DBAPISet` support both sets and frozensets for equality checks

JelleZijlstra:
The `__ne__` method was looking for sets and the `__eq__` method for frozensets.

Supporting both matches the behavior of `frozenset` itself.

### DIFF
--- a/pymysql/__init__.py
+++ b/pymysql/__init__.py
@@ -44,13 +44,13 @@ paramstyle = "pyformat"
 class DBAPISet(frozenset):
 
     def __ne__(self, other):
-        if isinstance(other, set):
+        if isinstance(other, (set, frozenset)):
             return frozenset.__ne__(self, other)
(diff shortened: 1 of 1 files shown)


methane:
Any real world usage?  I think they are garbage.
I hope no one use them.  No need for better compatibility with builtin types.

Just "Don't touch for backward compatibility."

methane:
For example, pymysql.STRING must be "type", not set, according to DB-API 2.0 spec.
https://www.python.org/dev/peps/pep-0249/#string

These globals violates spec.  I didn't touch them because anyone claim about it.
I had not seen any real world usage of them.

JelleZijlstra:
OK, I am not using them either—I just noticed that the code looks weird. For what it's worth, somebody using at least one of these (python/typeshed#2151).

PR #692, opened 2018-06-27
Use utf8mb4 for default encoding

methane:


henrykironde:
Is there a chance that this PR could have caused some Error with certain builds
https://github.com/appveyor/ci/issues/2505#issuecomment-403667339

core-code:
> Yes. If you used latin1 or other non-UTF8 encoding without explicitly setting it.

this change has definitely caused major issues for us. our webhost has updated PyMySQL without telling us, and then all communication with the database changed to another encoding. now we have 350.000 entries in the old charset and 3.500 entries in the new charset (most of which are duplicates of old entries with another charset) 

:(

> Use utf-8 always. Convert your DB.

can you recommend any way to convert the old entries to the new charset? is there any facility built into mysql 5.7.31 that could help us to convert the old entries? we are currently contemplating generating SQL statements to 'manually' update each affected old entry - is this the best course-of-action?

also, i guess specifying `charset='utf8mb4'` would be a good idea to prevent the charset suddenly changing again in the future?

thanks!

methane:
> can you recommend any way to convert the old entries to the new charset? is there any facility built into mysql 5.7.31 that could help us to convert the old entries? we are currently contemplating generating SQL statements to 'manually' update each affected old entry - is this the best course-of-action?

I am not MySQL admin expert. I am just MySQL protocol expert.
Please ask to MySQL Slack, or Stack Overflow.

PR #709, opened 2018-07-09
Add mysql native password plugin name on init handshake phase

elemount:
Hi methane,
Sorry for disturb you. We are using PyMySQL and have a private proxy code to handle mysql login phase. Recently we faced a connection problem when PyMySQL upgrade to 0.9, and after diagnostic we found that it is a proxy internal bug that assume the auth plugin name should not be empty. We fixed the bug on our hands, but we still want to add the auth plugin name to PyMySQL code like protocol. 

In the MySQL protocol(https://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::Handshake ), the client should figure out the auth_plugin_name
```
 if capabilities & CLIENT_PLUGIN_AUTH {
string[NUL]    auth plugin name
  }
```

We assume when the plugin is mysql_native_password, it should not be empty.

Thanks,
Shuode Li

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -805,6 +805,7 @@ class Connection(object):
         plugin_name = None
 
         if self._auth_plugin_name in ('', 'mysql_native_password'):
+            plugin_name = b'mysql_native_password'
             authresp = _auth.scramble_native_password(self.password, self.salt)
         elif self._auth_plugin_name == 'caching_sha2_password':
             plugin_name = b'caching_sha2_password'
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -804,7 +804,10 @@ class Connection(object):
         authresp = b''
         plugin_name = None
 
-        if self._auth_plugin_name in ('', 'mysql_native_password'):
+        if self._auth_plugin_name == '':
+            plugin_name = b''
+            authresp = _auth.scramble_native_password(self.password, self.salt)
+        elif self._auth_plugin_name == 'mysql_native_password':
             plugin_name = b'mysql_native_password'
             authresp = _auth.scramble_native_password(self.password, self.salt)
         elif self._auth_plugin_name == 'caching_sha2_password':


pymysql/connections.py
```diff
@@ -805,6 +805,7 @@ def _request_authentication(self):
         plugin_name = None
 
         if self._auth_plugin_name in ('', 'mysql_native_password'):
+            plugin_name = b'mysql_native_password'
             authresp = _auth.scramble_native_password(self.password, self.salt)
```
methane: When server sent `b''`, shouldn't we send `b''`?

elemount: You are right, let me fix it.

elemount: Hi methane, 
I do not think `When server sent b'', shouldn't we send b''?`. That is not correct.
The plugin_name should be consist with the `authresp` field, when response the plugin_name, it tells the server `authresp` method.

methane: Both of `b''` and `b'mysql_native_password'` are compatible with authresp, because both are same implementation.  Am I wrong?

>  The plugin_name should be consist with the authresp field, when response the plugin_name, it tells the server authresp method.

AFAIK, plugin_name in handshake response must be consistent with plugin_name server sent.  And authresp must consistent with plugin_name.  If plugin_name is not same to server sent, server may request `Auth Change`, without reading authresp.

Could you setup MySQL and MariaDB server which supports plugin, but send `b''` as plugin name on initial packet,
and capture all packets?
(Yes, developing auth flow is very bothersome job.  But even you think "it works", initial handshake may be broken and auth is succeeded after "Auth Switch")

elemount: Hi methane,
You are right. 
I think it too simple. After I tested MariaDB Embedded server, I find your knowledge and experience is correct. Thanks a lot for correcting me.

elemount:
@methane , I've fix the change after discussion. Could you help to review it?

PR #710, opened 2018-07-09
Fix old-style password logins

bmaranville:
During refactoring after May 7, 0.8.1 release support for old password logins was broken.  In _auth.py items were used without importing them, and in connections.py the function call to scramble_old_passwords was using the wrong arguments (needs self.salt)

### DIFF
--- a/pymysql/_auth.py
+++ b/pymysql/_auth.py
@@ -2,6 +2,7 @@
 Implements auth methods
 """
 from ._compat import text_type, PY2
+from .util import byte2int, int2byte
 from .constants import CLIENT
 from .err import OperationalError
 
@@ -12,6 +13,7 @@ from cryptography.hazmat.primitives.asymmetric import padding
 from functools import partial
 import hashlib
 import struct
+import io
 
 
 DEBUG = False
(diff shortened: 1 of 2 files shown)


methane:
When changing auth code, you need to test at least two scenarios:

* Initial handshake packet requires the auth method.
* Initial handshake packet requires other auth method.  The auth method is required by auth switch request.

bmaranville:
I admit I do not know the internals of the handshaking all that well - my testing was limited to:

- in version 0.8.1 and prior, I am able to connect to a MariaDB server which is using an old-style password with no errors.

- after version 0.8.1 I get an error, "bad handshake", for the same server, same username, same password

- looking at the code, I see that the call to scramble_old_passwords function was changed in part of the code when refactoring was done after version 0.8.1

- I changed it back to what it was in version 0.8.1, where self.salt is used as the second argument.

- now I am able to connect again.

________________________________

bmaranville:
I'm afraid I don't know how to test the two scenarios you describe - should I submit a ticket instead of a pull request?  The connection with old passwords is surely broken in versions > 0.8.1

methane:
I don't know too.  I need to study and try setup before test.
We OSS maintainers paying studying cost for setup we never use.

Since I don't have motivation about old_password, I don't have estimate
time to solve.
You can stick on old version of PyMySQL, but I strongly recommend
moving at least native_password.

PR #719, opened 2018-08-12
Convert SET field to python set

nenetchi:
Currently this library convert SET field to string or leave it as raw bytes (when use_unicode flag is set to False). However, this appears to be incorrect behaviour.

The C API documentation [states](https://dev.mysql.com/doc/refman/8.0/en/c-api-data-structures.html):
> ENUM and SET values are returned as strings. For these, check that the type value is MYSQL_TYPE_STRING and that the ENUM_FLAG or SET_FLAG flag is set in the flags value.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -17,7 +17,7 @@ import warnings
 from . import _auth
 
 from .charset import charset_by_name, charset_by_id
-from .constants import CLIENT, COMMAND, CR, FIELD_TYPE, SERVER_STATUS
+from .constants import CLIENT, COMMAND, CR, FIELD_TYPE, FLAG, SERVER_STATUS
 from . import converters
 from .cursors import Cursor
 from .optionfile import Parser
@@ -1232,11 +1232,15 @@ class MySQLResult(object):
                         encoding = None
                     else:
                         encoding = conn_encoding
+                    if field_type == FIELD_TYPE.STRING and field.flags & FLAG.SET:
+                        field_type = FIELD_TYPE.SET
                 else:
                     # Integers, Dates and Times, and other basic data is encoded in ascii
                     encoding = 'ascii'
             else:
                 encoding = None
+                if field_type == FIELD_TYPE.STRING and field.flags & FLAG.SET:
+                    field_type = FIELD_TYPE.SET
             converter = self.connection.decoders.get(field_type)
             if converter is converters.through:
                 converter = None


nenetchi:
There is another bug related to this PR:  
PyMySQL currently escape `set(['foo', 'bar'])` to `('foo','bar')`, and cause a `(1241, 'Operand should contain 1 column(s)')` error when assigned to a SET field.

In addition, 
1. changing type of return value may break some existing code. 
2. MySQL's SET is more similar to `enum.IntFlag`, is it appropriate to convert SET to python set?

Perhaps we should reevaluate this patch and ask for more opinions.

methane:
> There is another bug related to this PR:
> PyMySQL currently escape set(['foo', 'bar']) to ('foo','bar'), and cause a (1241, 'Operand should contain 1 column(s)') error when assigned to a SET field.
>
> In addition,
>
> changing type of return value may break some existing code.


It's not a bug.  It's design choice made by original author of MySQLdb,
and I won't break backward compatibility.

Rich type support "for human" can be implemented in ORM.
Changing it may break ORMs and many existing applications.


> MySQL's SET is more similar to enum.IntFlag, is it appropriate to convert SET to python set?
>
> Perhaps we should reevaluate this patch and ask for more opinions.
>

I'm interested only in FLAG_SET | TYPE_STRING, and TYPE_SET.
MySQLdb convert string to set only when TYPE_SET is used.

https://github.com/PyMySQL/mysqlclient-python/blob/3de469db63c5da47d61ac79bba5dadc0d22bde5c/MySQLdb/converters.py#L118

I don't know when MySQL returns FLAG_SET | TYPE_STRING and
TYPE_SET.  Maybe, we should return same value for them.  But I'm not sure.

nenetchi:
It seems that MySQL never return MYSQL_TYPE_SET.

L103 of [mysql-connector-c++](https://cdn.mysql.com/archives/mysql-connector-c++/mysql-connector-c++-1.1.9.tar.gz)`/driver/mysql_resultbind.cpp` says that
```
      // There two are not sent over the wire
    case MYSQL_TYPE_ENUM:
    case MYSQL_TYPE_SET:
```
and Oracle's `mysql-connector-python` convert TYPE_STRING | FLAG_SET to set even without checking for TYPE_SET.  https://github.com/mysql/mysql-connector-python/blob/master/src/mysql_capi.c#L2435

methane:
Thank you.
Then, MySQLdb's code had never worked.  Let's just remove it.

PR #720, opened 2018-08-13
Improve Connection's context manager

dash-rai:
A class's context manager is expected to return an instance of the class and close all resources when exiting. This PR attempts to standardize the behaviour of the Connection class's context manager.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -502,6 +502,7 @@ class Connection(object):
             self.rollback()
         else:
             self.commit()
+        self.close()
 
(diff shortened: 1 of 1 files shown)


methane:
It's not backward compatible.
We need deprecation period.

dash-rai:
I agree, but not calling `close()` in `__exit__` is dangerous enough to break compatibility in a major release.

methane:
I don't say I don't break backward compatibility.  I said we need deprecation period.

Additionally, current `__exit__` is for transaction.  Doing both of `close` and `commit` or `rollback` is unacceptable.  `commit` and `rollback` should be removed.

PR #721, opened 2018-08-22
If socket is not detected, attempt a reconnect

rubysown:
We ran into issues where after MySQL drops, any queries afterward would hit this exception and never be able to recover. This extra conditional statement helps correct that. You can reproduce this issue every time by leaving the extension running, restart MySQL and then continue attempting to use the program using the extension and you will receive the null InterfaceError statement. 

By adding this ping/reconnect we have solved the issue of the program (JupyterHub) becoming unavailable after a while.


methane:
This is low level library.  You need to use connection manager or pool over this,
instead of using connection directly.

PR #725, opened 2018-09-10
Cleanup after Python 2.6 support removed

adamchainz:
* Update `tox.ini` to match `.travis.yml` so Python 2.6 isn't listed there
* Use dictionary comprehensions where appropriate, a Python 2.7+ feature. They're a bit faster than creating a list and passing to `dict()`.


methane:
thanks

PR #726, opened 2018-09-11
[converter] use isinstance() instead of type() to get the encoder

deterralba:
Hi methane - and the other contributors, thank you for your work!

In the converters, the `type(val) == some type` comparison does not support children of python standard classes, it leads to tricky bugs (I use the `faker` library to generate random `datetime` object and it was failing because faker `datetime` are instance of a child class of `datetime`).

This PR changes this comparison and uses `isinstance(val, some type)` instead. 

NB: There is a tricky part: `datetime` inherit from `date`, hence we need to select the last child of the subclass chain. It is the role of `get_relevant_type`. The builds of the first commits were failing because of that (datetime was casted to date), it's working now.

### DIFF
--- a/pymysql/converters.py
+++ b/pymysql/converters.py
@@ -379,23 +379,27 @@ def get_encoder(val, mapping):
 
 
 encoders = {
+    text_type: escape_unicode,
+    # the order is important here: struct_time objects are also instances of tuple
+    # hence struct_time must come before tuple
+    time.struct_time: escape_struct_time,
     bool: escape_bool,
     int: escape_int,
     long_type: escape_int,
     float: escape_float,
     str: escape_str,
-    text_type: escape_unicode,
     tuple: escape_sequence,
     list: escape_sequence,
     set: escape_sequence,
     frozenset: escape_sequence,
     dict: escape_dict,
     type(None): escape_None,
-    datetime.date: escape_date,
+    # the order is important here: datetime objects are also instances of date
+    # hence datetime must come before date
     datetime.datetime: escape_datetime,
+    datetime.date: escape_date,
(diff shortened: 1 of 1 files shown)


deterralba:
I could not find any contributors guide lines, is there something else I should do?

grooverdan:
Can you add some tests to show the old/new behaviour is preserved?

deterralba:
Hi Daniel,

I think `test_basic.py:TestConversion:test_datatypes` already test all the type conversions, and ensure that the old behavior is the same for the types of `converters.encoders`. It was this test that was failing before I updated the code to select the last child of the subclass chain  - to use `datetime` instead of `date` when we have a subclass of `datetime`.

If you think we need a new test, I am unsure how I write it in a clean way:  I started to write something like
```
    def test_convert_isinstance_update(self):
        """Test that the use of isinstance work as expected for the common types."""
        from pymysql import converters
        types = [t for t in converters.encoders]
        for type_ in types:
            instance = create an instance from type_ # pseudo code
            assert get_relevant_type(instance, types) == type_
```
The step `create an instance from type_` would require an instantiation for all the keys of `converters.encoders,` it would not be very pleasant to write or maintain.

What are you thoughts on this?

methane:
I haven't reviewed yet this, but I will unlikely merge this type of pull
request.
My goal for this project is not user friendly high level API.

I focus on low level, stable, performance, and low maintenance cost.
This type of enhancement can be done in higher layer library like ORM.

>

methane:
This can be huge performance degrade.  I can not accept.
You can convert to datetime from subtype of it before passing it to PyMySQL.
You can register subtype to `conv`, although I don't like `conv` and I want to deprecate it in the future...

PR #744, opened 2018-11-12
Fix cryptography version in requirements.txt

montao:
I propose that we use a specific version instead of the very latest. cryptography has very recently released a new version (2.4)  ttps://github.com/pyca/cryptography/releases and this update breaks some builds in the following manner.:   

```
Downloading https://files.pythonhosted.org/packages/d2/5f/6ed3135eb1e775187f7ecd4e7713f1415516725365e51f9786143f36e024/cryptography-2.4.1.tar.gz (468kB)
  Could not find a version that satisfies the requirement cffi!=1.11.3,>=1.7 (from versions: )
No matching distribution found for cffi!=1.11.3,>=1.7
```

Therefore I propose to use a specific version here instead of using the "latest". It is also good practice to use a specific version instead of the latest. Do you agree? Thanks.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-cryptography
+cryptography==2.3.1
 


methane:
Why do you use `requirements.txt` in PyMySQL?
I added it only for local development *of* PyMySQL.
It is not used when *using* PyMySQL.

montao:
@methane If I install PyMySQL then the system will also install cryptography. On one system (our CI/CD pipeline) the upgrade of cryptography broke the build. Setting cryptography to previous version fixed it. Does that answer your question?

montao:
@montao I immediately did pin a specific version of cryptography in my project and the error went away and we got a clean build. I'm trying to understand why our build broke and help others who might experience something similar.

methane:
There are some reasons this is bad idea.

* If `cffi!=1.11.3,>=1.7` is really broken, cryptography's dependency should be fixed.
* But `cffi!=1.11.3,>=1.7` is not broken.  I can install cryptography==2.4.1 without any problem. (Of course, PyMySQL too).  So it's your environment specific issue.
* `requirements.txt` is not used when installing PyMySQL.  PyMySQL package doesn't contains the file.

I lock this issue because I don't want to pay my time and energy to teach you anymore.

Issue #750, opened 2018-12-13
ubyte format error at connection


davidshay:
When connecting from an app that starts in a folder deep down under root, I got the following error:
struct.error: ubyte format requires 0 <= number <= 255

Coming from this line of code in connections.py:
data += struct.pack('B', len(connect_attrs)) + connect_attrs

The workaround is to set the program_name at connection.

The fix would probably be to set program name using the file basename:
program_name = os.path.basename(argv[0])

PR #751, opened 2018-12-13
Avoid ubyte format error when file path is too long (#750)

davidshay:
I opened issue #750 for this problem.
When I open a connection from a file that has a very long path, I got a ubyte format error.


methane:
I prefer #755.

Issue #753, opened 2018-12-14
sys.argv call still results in an error


insideClaw:
The previous fix in PR #739, addressing bug #736, aims to handle sys.argv not being available, but it doesn't prevent an exception being thrown when it's not found. Exception extract shown below:
```
        File "/opt/datadog-agent/embedded/lib/python2.7/site-packages/pymysql/connections.py", line 321, in __init__
          argv = getattr(sys, "argv")
      AttributeError: 'module' object has no attribute 'argv'
```
The context is similar to the one in #736 - an environment that's unable to execute that system call. In my case the monitoring service Datadog using PyMySQL in its embedded environment.

insideClaw:
Fix implemented in #754

PR #754, opened 2018-12-14
Surround sys.argv call with a try-except block

insideClaw:
Fix for #753

Branched from master as the changes it's based on ( #739 ) have not gone through the development branch.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -315,12 +315,16 @@ class Connection(object):
             '_pid': str(os.getpid()),
             '_client_version': VERSION_STRING,
         }
-        if program_name is None:
-            argv = getattr(sys, "argv")
-            if argv:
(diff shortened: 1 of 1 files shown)


methane:
This code looks very ugly.  `except(AttributeError):` doesn't follow code style widely used for Python.
And `getattr(sys, "argv", None)` is enough for fixing this.

But sys.argv has another problem.  I'll remove automatically checking sys.argv.

methane:
* First of all, as I mentioned above, `getattr(sys, "argv", None)` is simplest.
* `except(AttributeError):` -> `except AttributeError:`
* try-except block must be small as possible; only `getattr` should be in try block.

insideClaw:
Thanks for explanation and the try-except block size optimization, I hadn't thought of that one.

PR #763, opened 2018-12-18
Remove context manager

methane:


ulidtko:
Woah, what an ample deprecation period! Whopping 7 weeks since #742.

*(stops wondering why our PyMySQL 0.8.0 doesn't provide context manager for Connection)*

methane:
PyMySQL 0.9.3 released at 2018-12-18.
So the deprecation period is more than 1 year.

PR #779, opened 2019-02-04
Optimize packet reading

Deneby67:
Optimized packet reading speed 26 times faster on my pc

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -638,7 +638,7 @@ class Connection(object):
         :raise OperationalError: If the connection to the MySQL server is lost.
         :raise InternalError: If the packet sequence number is wrong.
         """
-        buff = io.BytesIO()
+        buff = bytearray()
         while True:
             packet_header = self._read_bytes(4)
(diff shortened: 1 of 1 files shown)


Deneby67:
> What is your envirnonment?
> I thought `bytes +=` has optimization to avoid exponential execution time.

+= works str only

`Python 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0] on linux
`

`import time

def t1():
    t1 = time.time()
    b = b''
    for _ in range(1_000_000):
        b += b'i'
    print(f'concat {time.time() - t1}')

def t2():
    t1 = time.time()
    b = bytearray()
    for _ in range(1_000_000):
        b += b'i'
    print(f'byte array concat {time.time() - t1}')


def t3():
    t1 = time.time()
    b = bytearray()
    for _ in range(1_000_000):
        b.extend(b'i')
    print(f'byte array extend {time.time() - t1}')


for _ in range(5):
    t1()
    t2()
    t3()`

methane:
OK, thanks.

PR #780, opened 2019-02-13
Skip escape args if args is None or empty

JesseLivingston:


methane:
What is problem you want to solve?

PR #492, opened 2016-07-26
Birm 376

birm:
Referencing a feature request in #376. This is one of a handful of ways to handle file objects in addition to file paths for LoadLocalFile with minimal code and functionality change.

PR #782, opened 2019-02-28
Add interface for non-file LOAD DATA LOCAL INFILE (#376)

### LINKED ISSUE #376: LoadLocalFile should be able to take file like object

grooverdan:
I saw the comments in #492.

LOAD DATA LOCAL INFILE is a rapid load mechanism in MySQL. ETL, machine learning, and specialized accelerators are some of the reasons that a rapid generation of data can occur. These are
application domains where Python is frequently used. Placing generated data into to a file before a load is an additional complication. This commit provides an alternate to temporary file generation.

I have a WIP for mysql-python in the same way as this is implemented - https://github.com/grooverdan/mysqlclient-python/commit/e3cd1d85414fb3f893ec954cef7c2b178defb3a4

I n the same way as mysql-python, this implementation mirrors https://dev.mysql.com/doc/refman/5.7/en/mysql-set-local-infile-handler.html

The dictionary look up is equivalent to `local_infile_init` which set `*ptr` as the object returned for a given `filename`.

`local_infile_read` is equivalent to the `read` call on the object.

The `local_infile_end` can be handled by the object clean-up.

local_infile_error is equivalent to raising an exception.

More completely:

In this commit the load_infile argument is changed from a boolean to optionally a dictionary. In true pythonic style, a dictionary implements `__getitem__` and during the client side LOAD DATA LOCAL INFILE implementation it is the filename that is looked up in the dictionary.

The object retrieved from the dictionary will have have its `read` method called (argument max_packet_size) to retrieve data as if it was the file being uploaded. The `read` method will return CSV, or otherwise separated, data as bytes. When the `read` method is complete it will return None.

closes #376

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -156,9 +156,13 @@ class Connection(object):
     :param compress: Not supported
     :param named_pipe: Not supported
     :param autocommit: Autocommit mode. None means use server default. (default: False)
-    :param local_infile: Boolean to enable the use of LOAD DATA LOCAL command. (default: False)
+    :param local_infile: Boolean to enable the use of LOAD DATA LOCAL command.
+        Alternately, if a dictionary is specified, the filename of "LOAD DATA LOCAL INFILE 'name'"
+        will be looked up in the dictionary and the retreived file like object (has 'read' method)
+        that will be used to retrieve bytes (like a file) to send to the server.
+        (default: False)
     :param max_allowed_packet: Max size of packet sent to server in bytes. (default: 16MB)
-        Only used to limit size of "LOAD LOCAL INFILE" data packet smaller than default (16KB).
+        Only used to limit size of "LOAD DATA LOCAL INFILE" data packet smaller than default (16KB).
     :param defer_connect: Don't explicitly connect on contruction - wait for connect call.
         (default: False)
     :param auth_plugin_map: A dict of plugin names to a class that processes that plugin.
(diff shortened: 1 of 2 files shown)


grooverdan:
Alternate implementation that uses PEP-343 with statements:
https://github.com/PyMySQL/PyMySQL/compare/master...grooverdan:load_data_infile?expand=1

PROS:

For compatibility with a future mysql-python using [mysql_set_local_infile_handler()](https://dev.mysql.com/doc/refman/5.7/en/mysql-set-local-infile-handler.html), the `local_infile_error` error callback maps well to the callling the `__exit__` function of the iterated object.

CONS:

More boiler plate (two addition functions as per test case) to provide an implementation.

methane:
I don't like both API.  I prefer:

```py
db.load_data_infile(f)  # Executes LOAD LOCAL INFILE query and send data from f.
```

But please don't expect I am active on reviewing and merging PRs.
I am maintainer of some projects.  I develop new version of PyMySQL when I feel like it.
I don't know when, but it is not now.  Maybe, 6+ months later.

I'm very tired to manage PyMySQL 0.9.x and mysqlclient-1.4.x.
I don't want do anything other than important bugfix.

grooverdan:
Acknowledged. You've done really well over the years and thank you. I'll keep an eye out for mature python people who share your vision for a stable PyMySQL to lend a hand. I appreciate that even a shared maintainership isn't something you wish to consider at the moment.

Thank you for the API suggestion. I'll take a similarly deferred look at providing it. Keep well.

radyz:
Hey guys, did this feature got merged into master? I have a similar requirement where I need to pass in a file like object but it's actually a stream from google cloud storage. What I ended up doing was patching this class instead https://github.com/PyMySQL/PyMySQL/blob/6ccbecc1a0dfd04065b081950d2d35b1dac0aaa8/pymysql/connections.py#L1342

Issue #531, opened 2016-11-25
Exception in substitution sequence as parametr for Python 3


tonal:
## Current Behavior
```python
>>> curs.execute('SELECT 1 IN %s', ({1: 11, 2: 22, 3: 33}.keys(),))
...
AttributeError: 'dict_keys' object has no attribute 'translate'

>>> curs.execute('SELECT 1 IN %s', ({1: 11, 2: 22, 3: 33}.values(),))
...
AttributeError: 'dict_values' object has no attribute 'translate'
```
## Expected Behavior
```python
>>> curs.execute('SELECT 1 IN (1, 2, 3)')
1
>>> curs.execute('SELECT 1 IN %s', ([1, 2, 3],))
1
```

## Possible Solution
Add `dict_keys` and `dict_values` to pymysql.converters.encoders dictionary

## Executable script to reproduce (for bugs)

code:
```python
import pymysql.cursors

# Connect to the database
connection = pymysql.connect(host='localhost',
                             user='user',
                             password='passwd',
                             db='db')

try:
    with connection.cursor() as curs:
        curs.execute('SELECT 1 IN (1, 2, 3)')
        curs.execute('SELECT 1 IN %s', ([1, 2, 3],))
        curs.execute('SELECT 1 IN %s', ({1: 11, 2: 22, 3: 33}.keys(),)) # bug 1
        curs.execute('SELECT 1 IN %s', ({1: 11, 2: 22, 3: 33}.values(),)) # # bug 2
finally:
    connection.close()
```

schema:
any

## Tracebacks (for bugs)

For bug 1:
```
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/home/tonal/lang/projects/promsoft/epool/props_service/venv/lib/python3.5/site-packages/pymysql/cursors.py", line 164, in execute
    query = self.mogrify(query, args)
  File "/home/tonal/lang/projects/promsoft/epool/props_service/venv/lib/python3.5/site-packages/pymysql/cursors.py", line 143, in mogrify
    query = query % self._escape_args(args, conn)
  File "/home/tonal/lang/projects/promsoft/epool/props_service/venv/lib/python3.5/site-packages/pymysql/cursors.py", line 118, in _escape_args
    return tuple(conn.literal(arg) for arg in args)
  File "/home/tonal/lang/projects/promsoft/epool/props_service/venv/lib/python3.5/site-packages/pymysql/cursors.py", line 118, in <genexpr>
[...]

PR #787, opened 2019-03-22
If `val` is an iterable, try encoding it using the `list` encoder.

jfly:
This is hopefullly a generic solution to
https://github.com/PyMySQL/PyMySQL/issues/531 that doesn't come with any
downsides.

### DIFF
--- a/pymysql/converters.py
+++ b/pymysql/converters.py
@@ -1,5 +1,6 @@
 from ._compat import PY2, text_type, long_type, JYTHON, IRONPYTHON, unichr
 
+import collections
 import datetime
 from decimal import Decimal
 import re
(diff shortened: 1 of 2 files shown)


methane:
Do you know string is an iterator?

```
>>> import collections.abc
>>> isinstance("foo", collections.abc.Iterable)
True
```

I don't want complex heuristic rule.  Just convert it to list before passing it to pymysql.

jfly:
@methane, yes I was aware of that. I didn't think it's an issue because we still first look for the exact type match in `mapping` before we fall back to the `collections.Iterable` check. I did run the tests and they still passed with this change.

I understand that it's not terribly painful to just convert to a list before passing to pymysql, but this is a bit tricky when moving a codebase from Python 2 to Python 3, as all the places where we previously used `foo.keys()` or `foo.values()` need to be tweaked to call `list(foo.keys())` instead.

Issue #786, opened 2019-03-14
add ed22519 support


zzzeek:
Looking at issues like #651, #532, #583, it looks like pymysql does not wish to support any client side authentication modules, such as sha256, or ed25519 which is what MariaDB supports.    Two of the issues seem to be closed without any explanation though in #583 I see statements like ["I don't want to add dependency"](https://github.com/PyMySQL/PyMySQL/pull/583#issuecomment-310588369) ["Please move your code to pymysql/auth/sha256password.py"](https://github.com/PyMySQL/PyMySQL/pull/583#issuecomment-311917283).     

It doesn't seem like there is a "pymysql/auth" folder in the source tree.    

I have a need to understand what the path is here, my immediate use case is the ed25519 plugin.  So here are what I see as all the possible choices, you can check these off:

- [ ] PyMySQL supports pluggable authentication schemes via the `auth_plugin_map` parameter, and we welcome contributors to add these plugins to an as-yet-not-created folder `pymysql/auth`.   PyMySQL will distribute the plugins but not support them unless their authors do
- [ ] PyMySQL supports pluggable authentication schemes, but does *NOT* want any of the source code in the PyMySQL repo.   Please publish your plugins on pypi and maintain them separately.   (are there any such plugins ?  I couldn't find any)
- [ ] PyMySQL does not support authentication plugins at all, you've misunderstood what the `auth_plugin_map` argument does.  You will need to use mysql-connector-python or mysqlclient (does the latter support them?)
- [ ] other

methane:
I support sha256 and caching_sha256 already.
https://github.com/PyMySQL/PyMySQL/blob/master/pymysql/_auth.py

I never used pluggable auth.  I merged PR for adding auth_plugin_map with real experience.
While implementing sha256 and caching_sha256, I found current plugin API is not so good.
So I implemented them without plugin API.

methane:
BTW, I don't like cryptography dependency so I want to use pure Python solution.
I learned PKCS and sent pull request to python-rsa.  But it is not merged yet.

https://github.com/sybrenstuvel/python-rsa/pull/126

zzzeek:
> I support sha256 and caching_sha256 already.
> https://github.com/PyMySQL/PyMySQL/blob/master/pymysql/_auth.py
> 
> I never used pluggable auth. I merged PR for adding auth_plugin_map with real experience.
> While implementing sha256 and caching_sha256, I found current plugin API is not so good.
[...]

PR #791, opened 2019-04-17
Add ed25519 auth support

dciabrin:
Initial support for #786 based on PyNaCl. Currently, authentication is
limited to 32bytes-long passwords.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ matrix:
       python: "2.7"
     - env:
         - DB=mariadb:10.3
+        - TEST_MARIADB_AUTH=yes
       python: "3.7"
     - env:
         - DB=mysql:5.5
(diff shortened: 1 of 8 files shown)


dciabrin:
The test that I've added for ed25519 authentication has been run for the MySQL 8.0 build job. I'll rework the review to only trigger that test when using Mariadb 10.3+

dciabrin:
This new version is provides a full support for ed25519 auth mechanism, that depends on latest PyNaCl.
PyNaCl has committed https://github.com/pyca/pynacl/pull/528 which I'm using in this review, however they have not yet made a new release for it.

I've tried to configure travis to let pip download from git source, however this is slower and doesn't work with pip 9.0.1, which makes job #1327.3 fail.

Let's wait for a new PyNaCl release and I'll update this PR to "pip install PyNaCl" directly.

dciabrin:
PyNaCl 1.4.0 has just been released [1], so I'm rebasing the previous commit to consume it in CI.
Authenticating with auth_ed25519 works whatever the password length, and it looks like Travis is passing now.

[1] https://github.com/pyca/pynacl/issues/573#issuecomment-633722362

dciabrin:
@methane any chance you can have a look at this rebase? All automated tests are green now.

PR #800, opened 2019-06-25
Added Function Handling

14clarkt:
Copied and modified the method 'callproc' to handle functions. Named the resulting method as 'callfnct'. This was done for two reasons:
1) To make function calls that are safe from SQL Injections.
2) Remove the need to know the parameters required by the function by an equivalent (psudeo) execute call, for example:
cursor.execute("SELECT fnctname(_arg1, arg2, arg3, ..._)", args)
=>
cursor.callfnct(fnctname, args)

### DIFF
--- a/pymysql/cursors.py
+++ b/pymysql/cursors.py
@@ -270,6 +270,32 @@ class Cursor(object):
         self._executed = q
         return args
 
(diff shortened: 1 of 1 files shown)


methane:
I don't want this.

PR #820, opened 2019-11-06
More gracefully handles port coming in as string

brettl-sprint:
Currently, the user gets an unhelpful TypeError for a common and simple mistake. This should help point them in the right direction.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -276,6 +276,10 @@ class Connection(object):
         else:
             self.charset = DEFAULT_CHARSET
             self.use_unicode = False
+        
+        # Make sure port is an int.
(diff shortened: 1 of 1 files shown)


pymysql/connections.py
```diff
@@ -550,6 +550,9 @@ def set_charset(self, charset):
     def connect(self, sock=None):
         self._closed = False
         try:
+            # Make sure port is an int.
+            if not type(self.port) is int:
```
methane:
```suggestion
            if type(self.port) is not int:
```

methane:
Please move the check in `__init__`.

brettl-sprint:
> Please move the check in `__init__`.

Moved!

PR #822, opened 2019-11-07
Add Python 3.8 support

Djailla:
Python 3.8 is official now, test this package against this new version


.travis.yml
```diff
@@ -1,6 +1,5 @@
 # vim: sw=2 ts=2 sts=2 expandtab
 
-dist: xenial
```
methane: Don't remove this line.
I don't want to remember which distro is the default.

Djailla: Done

PR #830, opened 2019-12-04
Fix connection timeout error messages

svolle:
Fix inconsistency between connection read/write timeout error messages and actual value checks.


methane:
Thanks

PR #837, opened 2020-01-08
add support for using password for keyfile passed in via dictionary

carl-wallace:
add support for using password for keyfile passed in via sslp dictionary. works with load_cert_chain as described https://docs.python.org/3/library/ssl.html#ssl.SSLContext

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -336,11 +336,7 @@ class Connection(object):
         ctx.check_hostname = not hasnoca and sslp.get('check_hostname', True)
         ctx.verify_mode = ssl.CERT_NONE if hasnoca else ssl.CERT_REQUIRED
         if 'cert' in sslp:
-            password = sslp.get('password')
-            if password:
-                ctx.load_cert_chain(sslp['cert'], keyfile=sslp.get('key'), password=password)
-            else:
-                ctx.load_cert_chain(sslp['cert'], keyfile=sslp.get('key'))
+            ctx.load_cert_chain(sslp['cert'], keyfile=sslp.get('key'), password=sslp.get('password'))
         if 'cipher' in sslp:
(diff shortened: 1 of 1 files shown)


methane:
Would you provide a link of MySQL option?

methane:
I can not find MySQL support this. I don't think we need this although MySQL don't support it.

carl-wallace:
Impatient much? You ask for link, give no time to reply and then close (after letting this sit for months)? This is a mod that I am using, FWIW. Look at the API that is called. Why do you say not supported?

methane:
> Impatient much? You ask for link, give no time to reply and then close (after letting this sit for months)?

Sorry for bothering you.  I want to reformat code and close all pull requests before it.
So I searched the MySQL option by myself, instead of waiting reply. And I can not find any MySQL option.
If you have a link to MySQL option, please provide it?

>  FWIW. Look at the API that is called. Why do you say not supported?

I want to keep this library maintainable.  Rejecting requests not considered so useful is very important to keep the project maintainable.
And if it is so useful, why MySQL doesn't support it?  I usually rejects features MySQL doesn't support.

carl-wallace:
I have run into the problem with MySQL in a C/C++ library as well and made similar mods there (albeit not submitted as yet as a PR). It is useful because keeping plaintext private keys around is not a good practice. To turn your question around, the lower level API invoked by the mod I submitted *does* support it. Why would it support having a password on the P12 object if it were not useful?

methane:
I don't what "mod" means. I don't know how much is it useful. But if it is really useful for many users, it must be supported by MySQL Connector/C and mysql command.  So if you believe it is really useful very much, please request it to MySQL first.  If it is accepted, I will follow MySQL.

Anyway, PyMySQL supports passing SSLContext object.  You can use all ssl APIs already.

PR #840, opened 2020-02-14
Fix test suite compatibility with MySQL 8

javacruft:
MySQL 8 deprecates the use of display format for int columns:

 https://dev.mysql.com/doc/refman/8.0/en/numeric-type-syntax.html

This results in warnings being generated during test suite
execution which results in test failures.

Drop use of display widths - they don't materially change the tests
so this should be safe across all MySQL versions and variants.


methane:
Thanks.

PR #844, opened 2020-03-04
fix default section when using read_default_file

MrBenGriffin:
Closes #812


methane:
I don't want to use `[pymysql]` section by default.
Please support tuple for `read_default_group`.

methane:
This is MySQL Connector/C behavior.

> The client group is always read if you use MYSQL_READ_DEFAULT_FILE or MYSQL_READ_DEFAULT_GROUP.

Please follow it.

PR #846, opened 2020-03-21
Update to new issue template workflow

Margen67:
And more grammar fixes.


methane:
Thank you, but I used the default template instead.

PR #875, opened 2020-07-22
Add support for `cadata` SSL connection option in addition to `ca` and `capath`

jmehnle:
Add support for `cadata` SSL connection option in addition to `ca` and `capath`.

Per [`ssl.create_default_context`](https://docs.python.org/3/library/ssl.html#ssl.create_default_context) this allows the CA certificate to be passed in directly as a string rather than as a file or directory name.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -331,8 +331,9 @@ class Connection(object):
             return sslp
         ca = sslp.get('ca')
         capath = sslp.get('capath')
-        hasnoca = ca is None and capath is None
-        ctx = ssl.create_default_context(cafile=ca, capath=capath)
+        cadata = sslp.get('cadata')
+        hasnoca = ca is None and capath is None and cadata is None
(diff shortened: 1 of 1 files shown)


jmehnle:
I would've added a unit test, but I see no examples for testing SSL connection parameters and I didn't have time to dig all the way in. :( If you give me a hint, I may be able to write a unit test for this.

methane:
Please pass the SSLContext object, instead of adding more `ssl['xxx']` option.

All ssl options are bring from MySQL.  See this table.
https://dev.mysql.com/doc/refman/8.0/en/connection-options.html#encrypted-connection-options

`--ssl-ca` becomes `ssl['ca']`.
`--ssl-capath` becomes `ssl['capath']`.
`--ssl-cert` becomes `ssl['cert']`.

If you really want to add `ssl['cadata']`, please add it to MySQL first.  Then I will follow the MySQL.

jmehnle:
I didn't realize I was supposed to add an `SSLContext` object. Was there a hint in the docs I should've picked up on?

PR #886, opened 2020-08-17
Add context manager to Connection.

### LINKED ISSUE #735: With + Connection - returns cursor rather than the connection, doesn't close connection on exit
By returning a cursor rather than a connection, extra code needs to be written to handle exceptions and close the connection. Why not return a connection object, then a user can use a second nested with to get a cursor object. This way, wherever an error occurs, the cursors/connections will be neatly closed, and tracebacks will nicely propagate.

I noticed this was a problem when using SSH, and the tunnel wasn't getting decommissioned because the DB connection was still active. Other users also having this problem:

https://stackoverflow.com/questions/31214658/can-i-use-pymysql-connect-with-with-statement

Relevant lines in the code:

https://github.com/PyMySQL/PyMySQL/blob/09040abb0a8153b68b76113539632e7eaa53ce69/pymysql/connections.py#L494

https://github.com/PyMySQL/PyMySQL/blob/09040abb0a8153b68b76113539632e7eaa53ce69/pymysql/connections.py#L498

cmauro1:
Fixes #735.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -325,7 +325,14 @@ class Connection(object):
             self._sock = None
         else:
             self.connect()
-
+            
+    def __enter__(self):
+        return self
+    
+    def __exit__(self, *exc_info):
+        del exc_info
+        self.close()
+        
     def _create_ssl_ctx(self, sslp):
         if isinstance(sslp, ssl.SSLContext):
             return sslp


j08lue:
Why was the context manager functionality removed in the first place? Would you prefer a different way to use contexts with connection objects?

methane:
See #735.

methane:
I will merge this before releasing 1.0.
But current master branch is for 0.10.

j08lue:
I see, thank you.

PR #907, opened 2020-12-08
fix test unix_socket for MariaDB-10.4

grooverdan:
unix_socket is inbuilt in MariaDB-10.4+.

MariaDB-10.0 and 10.1 are out of support so replace these with MariaDB supported versions.

Like MySQL-5.5 keep MariaDB-5.5 as an oldest working version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 # vim: sw=2 ts=2 sts=2 expandtab
 
-dist: xenial
+dist: bionic
 language: python
 cache: pip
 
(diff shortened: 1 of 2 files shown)


grooverdan:
cryptography version bumped from 3.2.1 (https://travis-ci.com/github/PyMySQL/PyMySQL/jobs/456258876#L219) to 3.3 (https://travis-ci.com/github/PyMySQL/PyMySQL/jobs/456406912#L212) and failed.

grooverdan:
Thank you @methane

PR #930, opened 2021-01-05
Use keyword only argument

methane:


iamgd67:
wondering why change `connect` `host` arg oder, while this will break all code not using keyword but postion arguments.

methane:
Oh, that's my fault. I should made all arguments keyword-only.

iamgd67:
don't really understand what you mean by "all arguments keyword-only", I think lattest code still can pass position arguments by `connect('localhost','user','pass')`

methane:
> don't really understand what you mean by "all arguments keyword-only", I think lattest code still can pass position arguments by `connect('localhost','user','pass')`

Yes, and it's my fault. I had to forbid it when changing arguments order.

PR #937, opened 2021-01-07
tox.ini: Fix incorrect dependency syntax

mgorny:
Fix the dependency syntax in tox.ini.  The current syntax results in:

    ERROR: Invalid requirement: 'coverage pytest'


methane:
Thanks for reporting.
But I had not used tox for these years. I can not maintain it. It's time to remove it.

PR #954, opened 2021-01-24
Improve docstrings

darxriggs:
- dot at the end of descriptions
- 3rd instead of 2nd person
- more type information
- minor rephrasing

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -99,18 +99,18 @@ class Connection:
     Establish a connection to the MySQL database. Accepts several
     arguments:
 
-    :param host: Host where the database server is located
-    :param user: Username to log in as
+    :param host: Host where the database server is located.
+    :param user: Username to log in as.
     :param password: Password to use.
     :param database: Database to use, None to not use a particular one.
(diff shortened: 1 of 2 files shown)


pymysql/connections.py
```diff
@@ -99,18 +99,18 @@ class Connection:
...
+    :param host: Host where the database server is located.
+    :param user: Username to log in as.
     :param password: Password to use.
     :param database: Database to use, None to not use a particular one.
     :param port: MySQL port to use, default is usually OK. (default: 3306)
     :param bind_address: When the client has multiple network interfaces, specify
         the interface from which to connect to the host. Argument can be
         a hostname or an IP address.
-    :param unix_socket: Optionally, you can use a unix socket rather than TCP/IP.
+    :param unix_socket: Optionally, a unix socket rather than TCP/IP can be used.
```
methane, line 110: This sentence doesn't seem describing the `unix_socket` option.

Maybe, "Use a unix socket rather than TCP/IP." ?

darxriggs: I updated the text as suggested.

Issue #961, opened 2021-02-12
Connection.__del__ doesn't close the connection properly.


roelschroeven:
When you don't close a connection explicitly, the Python runtime eventually calls the `__del__` method which (IMO) should properly close the connection. PEP 249 says about Connection.close(): "Close the connection now (rather than whenever `.__del__()` is called)." That strongly suggests that the difference between `close()` and `__del__()` is not what happens, but only when it happens.

But `__del__` is bound to `_force_close`, which closes the connection without a QUIT message. MariaDB (and presumably MySQL) is not very happy with that. It writes warnings in its error.log:
`[Warning] Aborted connection 48939 to db: '****' user: '****' host: 'localhost' (Got an error reading communication packets)`

I could fix that in my client code of course, and I will. But still I think it would be better to do a proper close in `__del__`, if the connection is not already closed.

methane:
`__del__` is very danger method. Doing anything complex in it is bad idea. I want to minimize blocking operation in `__del__`.

roelschroeven:
That's a fair point.

PR #962, opened 2021-02-12
Close connection gracefully from Connection.__del__

roelschroeven:
Call close() from __del__() in Connection if the connection was not
already closed, to close the connection gracefully (i.e. first send
quite message to the server)

This solves issue #961

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -427,7 +427,11 @@ class Connection:
         self._sock = None
         self._rfile = None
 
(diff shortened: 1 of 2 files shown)


roelschroeven:
The pypy coverage test doesn't see any coverage in the __del__ method because its garbage collector works differently from the one in cpython. My changed __del__ method is not very useful in pypy, but I think it has use in cpython programs. I'm not sure what to do about that failed test.

PR #980, opened 2021-04-19
Add optional ssl_cadata arg

kukushking:
Hi there, 

Starting with Python 3.4 `ssl` module supports providing CA certs via [cadata](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_verify_locations). I suppose it's useful If you'd like to dynamically fetch the certificate from somewhere and avoid having to create a temporary file. 

This is exactly the use case in https://github.com/awslabs/aws-data-wrangler/pull/664 - we would like to fetch MySQL connection details from [AWS Glue Connections](https://docs.aws.amazon.com/glue/latest/dg/connection-using.html), including the CA certificate stored in AWS S3, if configured, and pass them over to PyMySQL.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -128,6 +128,9 @@ class Connection:
         (default: 10, min: 1, max: 31536000)
     :param ssl: A dict of arguments similar to mysql_ssl_set()'s parameters.
     :param ssl_ca: Path to the file that contains a PEM-formatted CA certificate.
+    :param ssl_cadata:
+        Either an ASCII string of one or more PEM-encoded certificates or
+        a bytes-like object of DER-encoded certificates.
     :param ssl_cert: Path to the file that contains a PEM-formatted client certificate.
     :param ssl_disabled: A boolean value that disables usage of TLS.
     :param ssl_key: Path to the file that contains a PEM-formatted private key for the client certificate.
@@ -193,6 +196,7 @@ class Connection:
         server_public_key=None,
         ssl=None,
         ssl_ca=None,
+        ssl_cadata=None,
         ssl_cert=None,
         ssl_disabled=None,
         ssl_key=None,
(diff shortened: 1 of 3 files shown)


kukushking:
@grooverdan Hi Dan - could you have a look, please?

kukushking:
Hi @lecram - Apologies for being annoying :) Could you have a look, please?

grooverdan:
@kukushking, recommend a little bit more patience. Some maintainers are exceptionally busy people.

On the code itself:
* Recommend squashing the commits so it look readable in the history
* Improve test cases to actually cover the use SSL functionality introduced. While the existing mock tests aren't great, taking the time to introduce a improved test case for your case, and the default path based SSL would be appreciated.

I assume your pinging of me and @lecram is based on previous contributions. As you can see, this was a considerable time ago and we're probably not the best people to review.

From https://github.com/awslabs/aws-data-wrangler/pull/664, it looks like you actually have a use case for this or potentially something you are working towards. Recommend stating the objective, because at the moment it looks like you are adding this just because you can. Maintainers, particularly busy ones, like to see a tangible case for reviewing and merging code.

kukushking:
@grooverdan Thanks for your feedback and apologies for any annoyance! Let me tag some of more recent contributors. @methane @darxriggs 
- Updated the PR description with use case.
- Squashed the commits (it is also possible during merge).
- There is an existing path-based [test case](https://github.com/PyMySQL/PyMySQL/blob/master/tests/test_auth.py#L24) and I added [cadata-based one](https://github.com/PyMySQL/PyMySQL/pull/980/files#diff-a99307c61c3c55875b5bb65e0b045ccd074e277afe3cca9665e7dbd33023daf9R31) as well. There are no SSL test cases for MariaDB, which is the reason for minor coverage decrease, however, default Docker image for MariaDB [isn't configured with SSL](https://github.com/PyMySQL/PyMySQL/blob/master/tests/test_mariadb_auth.py#L24) so so far I just had to go along with it. I'll see what I can do to improve coverage though.

methane:
I'm very conservative about adding more options.

Is the `ssl_cadata` a well-known option name for MySQL clients? Does `mysql` CLI have the option?
As far as I checked this list, I can not find ssl-cadata.
https://dev.mysql.com/doc/c-api/8.0/en/mysql-options.html

methane:
FYI, you can pass SSLContext object to `ssl=ctx` option. So you don't need to add this option nor use temporary file.
You have full control of SSLContext already.

kukushking:
Indeed, just passing `SSLContext` worked just fine. Seems I overlooked that option. Thanks for all help!

PR #988, opened 2021-06-25
Fix generating authentication response with long strings

netch80:
MySQL protocol uses special rules for string longer than 251 characters, listed here: https://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::HandshakeResponse
PyMySQL has already been using this for unpacking server messages, but was not used for forming "handshake response" from a client.
We have discovered this on failing of automated tests under Jenkins which are executed at very long paths. By default, PyMySQL fills `program_name` with full path of the running script. The current version generates single byte 0xfe as full length of "client connect attrs".
The proposed patch utilizes already present _lenenc_int() for prefixing formed strings.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -898,10 +898,10 @@ class Connection:
             connect_attrs = b""
             for k, v in self._connect_attrs.items():
                 k = k.encode("utf-8")
-                connect_attrs += struct.pack("B", len(k)) + k
+                connect_attrs += _lenenc_int(len(k)) + k
                 v = v.encode("utf-8")
-                connect_attrs += struct.pack("B", len(v)) + v
-            data += struct.pack("B", len(connect_attrs)) + connect_attrs
+                connect_attrs += _lenenc_int(len(v)) + v
+            data += _lenenc_int(len(connect_attrs)) + connect_attrs
 
         self.write_packet(data)
         auth_packet = self._read_packet()


pymysql/connections.py
```diff
@@ -879,10 +879,11 @@ def _request_authentication(self):
...
                 authresp = b"\0"  # empty password
 
-        if self.server_capabilities & CLIENT.PLUGIN_AUTH_LENENC_CLIENT_DATA:
+        if (
+            self.server_capabilities & CLIENT.PLUGIN_AUTH_LENENC_CLIENT_DATA
+            or self.server_capabilities & CLIENT.SECURE_CONNECTION
+        ):
             data += _lenenc_int(len(authresp)) + authresp
-        elif self.server_capabilities & CLIENT.SECURE_CONNECTION:
-            data += struct.pack("B", len(authresp)) + authresp
```
methane, line 885: Don't do this. See https://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::HandshakeResponse

```
  if capabilities & CLIENT_PLUGIN_AUTH_LENENC_CLIENT_DATA {
lenenc-int     length of auth-response
string[n]      auth-response
  } else if capabilities & CLIENT_SECURE_CONNECTION {
1              length of auth-response
string[n]      auth-response
```

"length of auth-response" is 1byte, not lenenc when `capabilities & CLIENT_PLUGIN_AUTH_LENENC_CLIENT_DATA` is not true.

netch80: Many excuses, Iʼve over-reacted without proper attention.
Iʼve pushed an updated variant which limits changes to the "client connection attributes" section.

Should I also add a test for this? The original issue can be triggered e.g. adding `program_name="x"*300` to connect() arguments (in this case, struct.pack() will fail), or cycle of connecting with program_name gradually increasing by 1 byte.

methane: If you can.  I love regression tests.  Otherwise, I may create a bug again.

Issue #992, opened 2021-07-27
Cursor is not an Iterator


sanchezg:
**Describe the bug**

Trying to iterate over a set of results: 

```
main_query = cursor.execute(QUERY, args=args)
result = next(main_query)
```

I get a:
`TypeError: 'DictCursor' object is not an iterator`

I think the problem is that `BaseCursor` [implements](https://github.com/PyMySQL/PyMySQL/blob/master/pymysql/cursors.py#L348) `__iter__` method and not the `__next__` method, as the Python reference states: https://docs.python.org/3/glossary.html#term-iterator

I made a workaround:

```
class DataBaseCursor(pymysql.cursors.DictCursor):
    def __iter__(self):
        return self

    def __next__(self):
        return next(iter(self.fetchone, None))
```

And it works as expected.

I can upload a PR with that for review, what do you think?

methane:
Currently, Cursor is an iterable, not an iterator. So missing `__next__` is not a bug. You can iterate over the cursor with `for` statement or `iter(cursor)`.

But DB-API 2.0 says `.next()` optional method. It is iterator protocol in Python 2. So having `__next__` makes sense to me.
Go ahead.

sanchezg:
Thanks for your response! And yes, you're right, my bad. In this case Cursor is an iterable. However I can work with having also `__next__` implemented.

methane:
Please create a PR.

sanchezg:
@methane  I couldn't find a contribution guidelines, I just upload a PR but please do not hesitate in point me if I need to follow a particular template or complete any additional docs or tests.

PR #995, opened 2021-08-03
Make Cursor an iterator

sanchezg:
Please check https://github.com/PyMySQL/PyMySQL/issues/992


pymysql/cursors.py
```diff
@@ -346,7 +346,10 @@ def _do_get_result(self):
         self._rows = result.rows
 
     def __iter__(self):
-        return iter(self.fetchone, None)
+        return self
+
+    def __next__(self):
+        return next(iter(self.fetchone, None))
```
methane, line 352:
```suggestion
        row = self.fetchone()
        if row is None:
            raise StopIteration
        return row
```

methane:
As LGTM.com reports, you need to fix SSCursor and SSDictCursor too.

PR #1008, opened 2021-09-21
keep the return type consistent whether there are results or not

qiugh:
when:
    records1 = cursor.fetchall()
    records2 = cursor.fetchall()  # len(records2) == 0
    records = records1 + records2
problem encountered:
    TypeError: can only concatenate list (not "tuple") to list

### DIFF
--- a/pymysql/cursors.py
+++ b/pymysql/cursors.py
@@ -298,7 +298,7 @@ class Cursor:
         """Fetch all the rows."""
         self._check_executed()
         if self._rows is None:
-            return ()
+            return []
         if self.rownumber:
             result = self._rows[self.rownumber :]
         else:


Nothing4You:
Per [DB API 2.0](https://www.python.org/dev/peps/pep-0249/#fetchall) it only requires `a sequence of sequences (e.g. a list of tuples)`.

https://dev.mysql.com/doc/connector-python/en/connector-python-api-mysqlcursor-fetchall.html suggests it `returns a list of tuples`, although I have not checked the code.

Currently, in PyMySQL the following types can be returned, not sure if I found all:

- empty Tuple from Cursor
https://github.com/PyMySQL/PyMySQL/blob/ee88d0f0e6499ad3054edbf057e08abfe25993c4/pymysql/cursors.py#L301
- Tuple from Cursor
https://github.com/PyMySQL/PyMySQL/blob/ee88d0f0e6499ad3054edbf057e08abfe25993c4/pymysql/connections.py#L1281
- List from SSCursor
https://github.com/PyMySQL/PyMySQL/blob/ee88d0f0e6499ad3054edbf057e08abfe25993c4/pymysql/cursors.py#L455

Nothing4You:
on another look, if i read that correctly, mysql.connector returns empty lists when it has no rows:
https://github.com/mysql/mysql-connector-python/blob/90eaeca65a6bbfc1fd9218aad5303957798215c3/lib/mysql/connector/cursor_cext.py#L501

and when it has results it can at least return
- rows
https://github.com/mysql/mysql-connector-python/blob/90eaeca65a6bbfc1fd9218aad5303957798215c3/lib/mysql/connector/connection.py#L798
- rows
https://github.com/mysql/mysql-connector-python/blob/90eaeca65a6bbfc1fd9218aad5303957798215c3/lib/mysql/connector/protocol.py#L517
- and rows
https://github.com/mysql/mysql-connector-python/blob/90eaeca65a6bbfc1fd9218aad5303957798215c3/lib/mysql/connector/protocol.py#L370

And for some reason GitHub doesn't like to embed code blocks for these links :(

methane:
I tried to change it long ago. But I can't do it because Django depended to the empty tuple.

I don't know current Django behavior. Please run Django test for all supported Django versions.

PR #1009, opened 2021-10-01
tests: container docker-entrypoint-initdb.d for ease of testing

grooverdan:
The goal is to make testing easier by contributors.

MySQL and MariaDB executable comments are used to preserve compatibility.

There was no way to use the docker-entrypoint-initdb.d as a volume in github actions as the service is started before the repository is cloned.

Moved MariaDB tests into test_connection.py.


methane:
thank you.

PR #1029, opened 2022-01-31
update pymysql.constants.CR

Nothing4You:
values from https://github.com/mysql/mysql-server/blob/mysql-8.0.28/include/errmsg.h

I noticed that #341 included `CR.CR_UNUSED_1`, but it was named `CR.CR_SECURE_AUTH` instead.
is that intentional?


methane:
> I noticed that #341 included `CR.CR_UNUSED_1`, but it was named `CR.CR_SECURE_AUTH` instead.
> is that intentional?


See this comment. https://github.com/PyMySQL/PyMySQL/pull/341#commitcomment-16842298

It was CR_UNUSED_1 when I generated the CR.py

Nothing4You:
> It was CR_UNUSED_1 when I generated the CR.py

It was and is `CR_UNUSED_1` in MySQL's `errmsg.h` but it was imported as `CR_SECURE_AUTH`, therefore not matching the name in `errmsg.h`.

PR #1043, opened 2022-04-06
Modify the default port parameter to MySQL's default port 3306

GxlHus:
As the title shows, modify the default port parameter to MySQL's default port 3306, this way you don't have to enter port=3306 every time.

### DIFF
--- a/pymysql/connections.py
+++ b/pymysql/connections.py
@@ -169,7 +169,7 @@ class Connection:
         host=None,
         database=None,
         unix_socket=None,
-        port=0,
+        port=3306,
         charset="",
         sql_mode=None,
         read_default_file=None,


methane:
PyMySQL need to distinguish the parameter is not specified or 3306 is explicitly specified.

GxlHus:
> PyMySQL need to distinguish the parameter is not specified or 3306 is explicitly specified.

But I think newbies would prefer to see the parameter port=3306 instead of None when looking at the help documentation in PyCharm

methane:
https://github.com/GxlHus/PyMySQL/blob/1c9324e4b496d322cbfcf69d72e06d76b54b568d/pymysql/connections.py#L106

This is enough.

Again, PyMySQL *need to distinguish* port=3306 and port is not specified.
This need is much more important than "newbies would prefer".

PR #1049, opened 2022-06-10
Update mariadb tests to 10.8, remove end of life mariadb 10.2

Nothing4You:


tgross35:
Out of curiosity - why are specific (and seemingly random)  combinations listed? E.g.  mdb10.7 + 3.11-dev, but 10.8 runs with 3.9

Seems like it would make sense to just matrix all possible combinations

```yaml
- db: ["mariadb:10.3", "mariadb:10.4", "mariadb:10.5", "mariadb:10.6", "mariadb:10.7", "mariadb:10.8"]
  py: ["3.7", "3.8", "3.9", "3.10", "3.11-dev"]
```

This isn't specific to this PR, just a general question

methane:
> Out of curiosity - why are specific (and seemingly random) combinations listed? E.g. mdb10.7 + 3.11-dev, but 10.8 runs with 3.9
> 
> Seems like it would make sense to just matrix all possible combinations outside of the `include` key, e.g.
> 

Because it is very unlikely there is a bug happen only in specific combination of Python version and DB version.

I don't waste energy and computing resources because:

* MS helps OSS a lot, but it is limited resource. We are free riding on it.
* Using some energy cause:
  * Emits some CO2
  * Keep energy price high - keep fuel price high - keep Russian invation

Every 1Wh is 1ml of blood.

PR #1056, opened 2022-07-10
Expose `Cursor.warning_count`

Nothing4You:
In #774 automatic warnings were removed.
As far as I can tell there is currently no public API to access `cursor._result.warning_count`.

This provides a way to check for existence of warnings without having to perform an additional query over the network.


LostInDarkMath:
This is exactly what I needed :)

PR #1077, opened 2022-11-18
Connection pool feture

PyRSA:
### Add new function of connection pool

- Use connection pool
```python
import pymysql.cursors
import pymysql.pool

# connect config
mysql_settings = dict(
    host='localhost',
    user='user',
    password='passwd',
    database='db',
    cursorclass=pymysql.cursors.DictCursor
)

## Create a ConnectionPool object

pool = pymysql.pool.ConnectionPool(min_idle=1, max_idle=3, **mysql_settings)

## Get a connection object from the connection pool
# When the connection is used up, it will be automatically put back into the connection pool
# instead of closing the connection.

# * If you use the `with` statement, you don’t need to manually call the `close()` method.
#   When the connection object exits after use, it will be called automatically.
# * If you do not use the `with` statement, you need to manually call the `close()` method of the
#   connection after using the connection, or call the `return_connection(connection)` method of
#   the connection pool to return the connection to the connection pool.

connection = pool.get_connection()

with connection:      
    with connection.cursor() as cursor:  
        # Create a new Table
        create_table_sql = """
        CREATE TABLE `users` (
               `id` int(11) NOT NULL AUTO_INCREMENT,
               `email` varchar(255) COLLATE utf8_bin NOT NULL,
               `password` varchar(255) COLLATE utf8_bin NOT NULL,
               PRIMARY KEY (`id`)
           ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
           AUTO_INCREMENT=1 ;
        """
        cursor.execute(create_table_sql)
        
        # Create a new record
        sql = "INSERT INTO `users` (`email`, `password`) VALUES (%s, %s)"
        cursor.execute(sql, ('webmaster@python.org', 'very-secret'))

    # connection is not autocommit by default. So you must commit to save
    # your changes.
    connection.commit()

    with connection.cursor() as cursor:
        # Read a single record
        sql = "SELECT `id`, `password` FROM `users` WHERE `email`=%s"
        cursor.execute(sql, ('webmaster@python.org',))
        result = cursor.fetchone()
        print(result)
```

- Use connection pool (multithreading)
```python
import threading
import pymysql.pool

# connect config
mysql_settings = dict(
    host='localhost',
    user='user',
    password='passwd',
    database='db',
    cursorclass=pymysql.cursors.DictCursor
)

## Create a ConnectionPool object

pool = pymysql.pool.ConnectionPool(min_idle=1, max_idle=3, **mysql_settings)


def opera(conn_pool):
    tn = threading.current_thread().name
    conn = conn_pool.get_connection()
    cur = conn.cursor()
    cur.execute("show tables;")
    print(f"Thread: {tn}, connection: {conn}, result: {cur.fetchall()}")
    cur.close()
    conn.close()


threads = [threading.Thread(target=opera, args=(pool,)) for _ in range(5)]
for thread in threads:
    thread.start()
for thread in threads:
    thread.join()
print('pool size:', pool.size, 'pool created connect num', pool.connect_num)
```

- For more examples, please refer to the test case: `pymysql/tests/test_pool.py`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -114,6 +114,7 @@ The following examples make use of a simple table
 
 
 .. code:: python
+
     ## Use Connection Pool
 
     import pymysql.cursors
--- a/README.rst
+++ b/README.rst
@@ -118,6 +118,7 @@ The following examples make use of a simple table
     ## Use Connection Pool
 
     import pymysql.cursors
+    import pymysql.pool
 
     # connect config
     mysql_settings = dict(
--- a/README.rst
+++ b/README.rst
@@ -113,6 +113,53 @@ The following examples make use of a simple table
             print(result)
 
 
+.. code:: python
+    ## Use Connection Pool
+
+    import pymysql.cursors
+
+    # connect config
+    mysql_settings = dict(
+        host='localhost',
+        user='user',
+        password='passwd',
+        database='db',
+        cursorclass=pymysql.cursors.DictCursor
+    )
+
+    ## Create a ConnectionPool object
+    pool = pymysql.pool.ConnectionPool(min_idle=1, max_idle=3, **mysql_settings)
+
+    ## Get a connection object from the connection pool
+    # When the connection is used up, it will be automatically put back into the connection pool
+    # instead of closing the connection.
+
+    # * If you use the `with` statement, you don’t need to manually call the `close()` method.
+    #   When the connection object exits after use, it will be called automatically.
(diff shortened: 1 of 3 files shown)


methane:
I don't want to maintain pool.
I recommend people to use SQLAlchemy engine instead. It is battle-tested, connection pool.

PR #1090, opened 2023-02-18
Add fetch_table_names

### LINKED ISSUE #789: FEATURE REQUEST: Add ability to include table name for all returned column names/values
This feature request is implemented to a certain extent when using the DictCursor class to prevent name collisions. [here](https://github.com/PyMySQL/PyMySQL/blob/3674bc6fd064bf88524e839c07690e8c35223709/pymysql/cursors.py#L370)
However, I would like to see an option similar to what PHP 7 has like this:
`$pdo->setAttribute(PDO::ATTR_FETCH_TABLE_NAMES, 1);`
I just want a way to do something like so:
```
    db = pymysql.connect(
        host=db_host, 
        db=db_name, 
        port=db_port, 
        user=db_user, 
        password=db_password,
        cursorclass=pymysql.cursors.DictCursor,
       fetch_table_names=true
    )
```

Niecke:
This adds a parameter fetch_table_names to the connection object. When set to true, the table name is always prepended to the column names.

Fixes #789


methane:
This should be option of DictCursor, not Connection.

methane (changes requested):
This should be option of DictCursor, not Connection.

PR #1133, opened 2023-08-22
feat add charset

sean-k1:
It seems that the information about charset-related details based on the ID is not up to date. Therefore, I added the information according to what is written in the charset.py generate script.

use this script
```
mysql -N -s -e "select id, character_set_name, collation_name, is_default
from information_schema.collations order by id;" | python -c "import sys
for l in sys.stdin.readlines():
    id, name, collation, is_default = l.split(chr(9))
    if is_default.strip() == 'Yes':
        print('_charsets.add(Charset(%s, \"%s\", \"%s\", True))' \
              % (id, name, collation))
    else:
        print('_charsets.add(Charset(%s, \"%s\", \"%s\"))' \
              % (id, name, collation))"
```


methane:
I manually strip collation id larger than 255.
Collation id is used for handshake, and collatin id in handshake packet is only 1 byte.
So we can not use collation id larger than 255.

PR #1145, opened 2023-11-15
Added ssl_key_password param

svaskov:
Added support for SSL private key password in Connection class to handle encrypted keys.


svaskov:
I couldn't manage to handle Django tests because I can't understand what exactly goes wrong, it would be nice if someone pointed me in the right direction of thought.

methane:
Don't care about Django tests. It is expected failure.

svaskov:
@methane Thanks a lot for your help and for all your efforts you put into the project!

PR #1163, opened 2024-02-14
Allow PyMySQL to use encoding autodetection

rgryta:
This PR adds a new optional flag that enables usage of `charset-normalizer` package to detect the output encoding.

If this feature is accepted, I'll add some unit tests of course.


methane:
I don't like it. MySQL has connection encoding. We should use it instead of guessing.

PR #1165, opened 2024-03-26
test json - mariadb without JSON type

grooverdan:
MariaDB-11.0.1 removed the 5.5.5 version hack (MDEV-28910).

MariaDB still doesn't support JSON as a type.

Use get_mysql_vendor() == mysql for the final part of test_json.

As tested: https://buildbot.mariadb.org/#/builders/158/builds/7720/steps/4/logs/stdio


methane:
thanks

PR #1184, opened 2024-09-03
make sure to also close `connection._rfile` in `Connection._force_close`

### LINKED ISSUE #1183: `Connection._force_close` relies on garbage collector to close `SocketIO` instances
**Describe the bug**
We've had a bug report in PyPy that sockets aren't closed correctly when using `PyMySQL`: https://github.com/pypy/pypy/issues/5020
I've have looked into it a bit, it seems that `Connection._force_close` does not close the `self._rfile` attributes before setting them to `None`. Since socket objects have a manual refcount how many times `.makefile` is called on them, this means that the socket file descriptor isn't closed immediately.

In CPython this isn't visible due to reference counting closing the `SocketIO` instance in a timely manner. But in PyPy, this could happen only at the end of the next major garbage collection cycle, which could be quite a bit later.

**To Reproduce**
Complete steps to reproduce the behavior:

```python
import sys
import os
import pymysql

db_config = {
    'host': 'localhost',
    'user': 'rob',
    'password': '123456',
    'connect_timeout': 1,
    'read_timeout': 1,
    'write_timeout': 1,
}

def netstat():
    os.system(f"netstat -anp 2>/dev/null | grep 3306 | grep {os.getpid()}")

connection = pymysql.connect(**db_config)


connection = pymysql.connect(**db_config)
# keep the socket alive to simulate the slowness in calling __del__ of pypy
sock = connection._sock
rfile = connection._rfile

print(f"{connection=} {connection._sock=} {connection._rfile=}")
netstat()
connection._force_close()
print(f"after _force_close {connection=} {connection._sock=} {connection._rfile=}")
print(f"{sock=} {rfile=}")
netstat()
```

**Expected behavior**
I would expect the final `netstat()` call to not show an open socket.

**Environment**
 - OS: Linux
 - Server and version:  5.7.41
 - PyMySQL version: 1.4.6

**Additional context**
I'm happy to try to open a drive-by PR for this.

cfbolz:
this should fix #1183.

### DIFF
--- a/pymysql/tests/test_connection.py
+++ b/pymysql/tests/test_connection.py
@@ -909,6 +909,7 @@ class TestEscape(base.PyMySQLTestCase):
             rfile = con._rfile
             assert rfile is not None
             raise TypeError
+
         con._request_authentication = _request_authentication
 
         with pytest.raises(TypeError):
--- a/pymysql/tests/test_connection.py
+++ b/pymysql/tests/test_connection.py
@@ -894,3 +894,28 @@ class TestEscape(base.PyMySQLTestCase):
         assert rfile.closed
         assert sock._closed
         assert sock.fileno() != fileno  # should be set to -1
+
+    def test_socket_closed_on_exception_in_connect(self):
+        con = self.connect(defer_connect=True)
+        sock = None
+        rfile = None
+        fileno = -1
+
+        def _request_authentication():
+            nonlocal sock, rfile, fileno
+            sock = con._sock
+            assert sock is not None
+            fileno = sock.fileno()
+            rfile = con._rfile
+            assert rfile is not None
+            raise TypeError
+        con._request_authentication = _request_authentication
+
+        with pytest.raises(TypeError):
+            con.connect()
+        assert not con.open
+        assert con._rfile is None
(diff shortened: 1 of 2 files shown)


pymysql/tests/test_connection.py
```diff
@@ -883,3 +883,14 @@ def test_commit_during_multi_result(self):
...
+    def test_force_close_closes_socketio(self):
+        con = self.connect()
+        sock = con._sock
+        fileno = sock.fileno()
+        rfile = con._rfile
+
+        con._force_close()
+        assert rfile.closed
+        assert sock._closed
+        assert sock.fileno() != fileno # should be set to -1
```
methane, line 896:
```suggestion
        assert sock.fileno() != fileno  # should be set to -1
```

cfbolz:
@methane thanks for enabling the CI runs and fixing the spacing!

The complaint of coverage is that it seems the catch-all `except BaseException` in `Connection.connect` isn't being hit by any test, right?

I'll try to add a test for that. Also, the code in that except is basically `_force_close` now, maybe we should simply call that there.

methane:
thanks

Issue #1180, opened 2024-08-19
MariaDB in repo description


robertsilen:
Hello, could somebody add MariaDB to the repo description?  Similar to what was done for PyMySQL/mysqlclient in https://github.com/PyMySQL/mysqlclient/issues/719 ? Cheers

theneuralcraftsman:
I updated the repo description.

theneuralcraftsman:
PR (#1186) was merged.

PR #1185, opened 2024-09-16
Add MariaDB mention in repo description and improve example with exception handling (#1180)

theneuralcraftsman:
This PR addresses issue #1180, which requested adding MariaDB to the repository description, similar to how it was done for PyMySQL/mysqlclient in PyMySQL/mysqlclient#719.

Additionally, I updated the example code by incorporating try, except, and finally blocks to ensure proper error handling and cleanup when interacting with the database. This improves the robustness of the code and ensures that the connection is always closed, even in case of an error.

## Changes made:
- Added MariaDB to the repo description.
- Updated the example with error handling and resource management using try, except, and finally.

Let me know if further changes are needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 # PyMySQL
 
-This package contains a pure-Python MySQL client library, based on [PEP
+This package contains a pure-Python MySQL and MariaDB client library, based on [PEP
 249](https://www.python.org/dev/peps/pep-0249/).
 
 ## Requirements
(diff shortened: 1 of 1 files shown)


methane:
It reduces the robustness.

- remove traceback from the error.
- `with` statement ensures that connections closed always already.

You should avoid try statement as possible to write robust code.
Read the "A Philosophy of Software Design, 2nd Edition".

theneuralcraftsman:
Thanks for the feedback and the book recommendation! I've reverted the try-except changes and updated the repository description to include MariaDB as requested in PR(#1186).

PR #1223, opened 2026-02-26
fix: replace 1 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.


methane:
This is not a bug fix. Rather, it is causing a bug.

PR #1249, opened 2026-05-29
security: restrict LOAD DATA LOCAL INFILE to query arguments via local_infile='only_args'

dfgvaetyj3456356-hash:
Revised fix addressing maintainer feedback from #1247. The new local_infile='only_args' mode restricts file loading to paths explicitly passed as query arguments, preventing arbitrary file reads.


methane:
I reject this idea.

PR #1266, opened 2026-08-30
test: add reproduction test for callproc multi-resultsets with OUT params

soepic1:
## Summary

Adds a dedicated test case to verify `cursor.callproc()` behavior when executing stored procedures that return internal result sets alongside `OUT` parameters.

## Details

When a MySQL stored procedure contains internal queries (e.g., `SELECT` statements) and updates `OUT` parameters, the server returns multiple result set packets over the socket. 

This test ensures that:
1. Internal `SELECT` result sets remain accessible via `cursor.fetchall()`.
2. Draining result sets using `cursor.nextset()` allows session variables (`@_procname_n`) to be fetched reliably via subsequent `SELECT` queries without socket desynchronization.

## Testing

- Added `test_callproc_with_multi_resultset_and_out_params` under `pymysql/tests/test_issues.py`.
- Verified passing status locally against MySQL 8.0 / Cloud SQL proxy.
- Ran full integration test suite (`pytest pymysql/tests/`) to ensure zero regressions across standard `callproc` usage.

### DIFF
--- a/pymysql/tests/test_issues.py
+++ b/pymysql/tests/test_issues.py
@@ -518,3 +518,35 @@ class TestGitHubIssues(base.PyMySQLTestCase):
             self.assertEqual(cur.fetchone()[0], 246)
         finally:
             cur.execute("DROP PROCEDURE IF EXISTS `foo.bar`")
+
+class TestCallProcEdgeCases(base.PyMySQLTestCase):
+    def test_callproc_with_multi_resultset_and_out_params(self):
+        conn = self.connections[0]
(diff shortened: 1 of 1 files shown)


soepic1:
addressed Copilot feedback by adding try/finally cleanup and explicit nextset() assertions.

soepic1:
Hi @PyMySQL maintainers, just checking in on this when you have a moment! 

All CI checks are passing cleanly. Please let me know if you would like any adjustments to the implementation or additional test cases added. Happy to make updates!

